PR #9, opened 2020-02-14
LoRaWAN packet routing (HIP draft)

Vagabond:
Rendered view: https://github.com/helium/HIP/blob/c2f3ce61466b003731bb967959ca8b6e7706fca5/0008-lorawan-routing.md

### DIFF
--- a/0008-lorawan-join-routing.md
+++ b/0008-lorawan-join-routing.md
@@ -47,9 +47,10 @@ fastest hash functions available and seems to provide good distribution for the
 XOR filter. Both the DevEUI and the AppEUI (totaling 128 bits together) would be
 hashed to a 64 bit key, this should help in cases of DevEUI collisions.
 
-A new ledger entity would be added to the blockchain, a 'routing entry'. These
-would have a total ordering (assigned by the order they are added to the chain)
-and an XOR filter built from the set of active DevEUI/AppEUIs for the owning OUI.
+A list of XOR filters would be attached to the OUI record in the ledger that
+would contain the routing filters for that OUI. A single filter would suffice
+for small OUIs, only larger OUIs whose amount of devices could exceed the
+maximum filter size would require more than one routing filter.
 
 XOR16 filters for a range of sizes are presented below for illustration purposes:
 
@@ -203,6 +204,12 @@ generate the routing tables, run a second time to use the generated routing
 tables, and then the tables were copied to the Pi to run the test with the same
 databases (generating the tables can be quite RAM intensive, so it is unsuitable
 to be done on the Pi).
+
+The process of generating the filter is roughly as follows: the complete list of
+{DevEUI, AppEUI} pairs are hashed into a 64 bit keys and then fed into the XOR
+filter generator. The filter then does some bit twiddling to set some bits in a
+list of 'fingerprints'. A fairly readable example can be found
+[here](https://github.com/juanbono/xor-filter/blob/master/src/lib.rs).
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/0008-lorawan-join-routing.md
+++ b/0008-lorawan-join-routing.md
@@ -9,8 +9,8 @@ Many LoRaWAN sensors come pre-provisioned and as such AppEUI and AppKey cannot b
 We would like to implement a scheme that takes AppEUI, AppKey, and DevEUI as fixed parameters, but still enables hotspots to route data to the proper endpoint.
 This HIP proposes to add [XOR filter](https://lemire.me/blog/2019/12/19/xor-filters-faster-and-smaller-than-bloom-filters/)
 based routing tables to the blockchain ledger to aid in routing LoRaWAN join
-packets to the correct destination. Additionally once the device has joined, it
-can be assigned a DevAddr corresponding to an address allocation the OUI controls.
+packets to the correct destination. A related HIP describes the routing of
+packets once the device has joined.
 
 # Motivation
 [motivation]: #motivation
@@ -24,12 +24,6 @@ is much closer to internet routing. Additionally, it turns out many LoRaWAN devi
 do not allow end-user configuration of DevEUI/AppEUI/AppKey. Thus the existing
 model of co-opting the AppEUI to route the join packet is unsuitable and
 something new is needed.
-
-In addition, the current scheme for DevAddr allocation (the session address,
-essentially) is not ideal because we map the 32 bit OUI to the DevAddr directly.
-This will cause problems for large OUIs at the expense of small ones. It would
-be good for OUIs to be able to obtain more address space for their DevAddrs if
-they're willing to pay for it.
 
 # Stakeholders
 [stakeholders]: #stakeholders
@@ -51,7 +45,7 @@ compared against the XOR filter. We propose the use of
 [xxhash](http://cyan4973.github.io/xxHash/)'s 64 bit mode. This is one of the
 fastest hash functions available and seems to provide good distribution for the
 XOR filter. Both the DevEUI and the AppEUI (totaling 128 bits together) would be
-hashed to a 64 bit key, this should help in cases of DevAddr collisions.
+hashed to a 64 bit key, this should help in cases of DevEUI collisions.
 
 A new ledger entity would be added to the blockchain, a 'routing entry'. These
 would have a total ordering (assigned by the order they are added to the chain)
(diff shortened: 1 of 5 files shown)


amirhaleem:
> This seems like a very nice solution to the problem of on-boarding devices with arbitrary credentials. While I believe we should focus on this routing scheme in the short-term, I would note that we have two paths to reduce Join latencies should that become a friction point for network usage:
> 
> * we could use a reserved [DevEUI](https://lora-developers.semtech.com/library/tech-papers-and-guides/the-book/deveui/) range to be a routing "routing mode" indicator; this would allow you to use part of the DevEUI or the AppEUI as a way to signal OUI/deviceID
> * we could use the MType bit field to make a whole different message, either Join or normal message

We should be cognizant that no off-the-shelf LoRaWAN devices will work this way - we ideally want a solution that is compatible with any LoRaWAN device, not requiring any changes to their credentials or firmware. I'd argue overloading DevEUI is not much of an improvement over the current AppEUI hack we are using today.

0008-lorawan-routing.md
```diff
@@ -0,0 +1,301 @@
...
+We will be unable to support routing join packets for arbitrary LoRaWAN
+join packets and thus unable to support LoRaWAN devices that are
+already provisioned or unable to have their credentials modified.
+
+# Unresolved Questions
+[unresolved]: #unresolved-questions
+
+## What parts of the design do you expect to resolve through the HIP process before this gets merged?
+
+We need to nail down the permissible sizes of the XOR filters we want to store.
```
amirhaleem: I think you will have to propose something here @Vagabond, I am not sure how to evaluate this.

0008-lorawan-routing.md
```diff
@@ -0,0 +1,301 @@
...
+
+We need to decide if these routing table entries also contain a range of
+DevAddr entries to be used for routing of non-join packets.
```
amirhaleem: What are the tradeoffs here? If we don't pre-assign DevAddr's to OUI's, how would a Hotspot lookup the correct OUI on subsequent non-join uplinks?

0008-lorawan-routing.md
```diff
@@ -0,0 +1,301 @@
...
+
+## What parts of the design do you expect to resolve through the implementation of this feature?
+
+We need to define where/how these routing entries are stored, how they're
+created/updated and if they can be transferred (like IP space can be traded).
+
+## What related issues do you consider out of scope for this HIP that could be addressed in the future independently of the solution that comes out of this HIP?
+
+Pricing and availability of routing table entries for organizations operating
```
amirhaleem: The economics are out of scope here - will kick off a separate discussion/HIP on this.

0008-lorawan-routing.md
```diff
@@ -0,0 +1,301 @@
...
+## What parts of the design do you expect to resolve through the HIP process before this gets merged?
+
+We need to nail down the permissible sizes of the XOR filters we want to store.
+
+We need to decide if these routing table entries also contain a range of
+DevAddr entries to be used for routing of non-join packets.
+
+## What parts of the design do you expect to resolve through the implementation of this feature?
+
+We need to define where/how these routing entries are stored, how they're
```
amirhaleem: Similar to earlier comment I think you should propose a solution (or just start building for future review) - I'm assuming you are talking at the ledger/rocksdb level, and what the transaction structure should look like?

lthiery:
> We should be cognizant that no off-the-shelf LoRaWAN devices will work this way - we ideally want a solution that is compatible with any LoRaWAN device, not requiring any changes to their credentials or firmware. I'd argue overloading DevEUI is not much of an improvement over the current AppEUI hack we are using today.

Right - the proposed solution here is great because it allows for arbitrary credentials and creates universal compatibility for off-the-shelf LoRaWAN devices. 

I'm just leaving notes on a few ideas on how to reduce latencies while remaining compatible with this proposed scheme; I am not suggesting either of those routes be considered for development today.

0008-lorawan-routing.md
Vagabond: I have added a second HIP to this PR that addresses DevAddr routing more thoroughly

0008-lorawan-join-routing.md
```diff
@@ -0,0 +1,294 @@
...
+Average lookup 0.916s, max 1.009s, min 0.864s
+Lookup misses 0
+```
+
+The benchmark code can be found
+[here](https://github.com/Vagabond/bloom_router). These runs were generated by
+the command in the README. The benchmark was first run on the Intel machine to
+generate the routing tables, run a second time to use the generated routing
+tables, and then the tables were copied to the Pi to run the test with the same
+databases (generating the tables can be quite RAM intensive, so it is unsuitable
```
lthiery: How will tables be updated then? Every time someone registers a set of IDs, someone has to do the work to generate and distribute the new table?

Vagabond: Yes, so you'd likely want to batch updates.

0009-lorawan-devaddr-routing.md
```diff
@@ -0,0 +1,173 @@
...
+- HIP PR: <!-- leave this empty -->
+- Tracking Issue: <!-- leave this empty -->
+
+# Summary
+[summary]: #summary
+
+This HIP proposes a routing scheme for LoRaWAN packets post-join, via DevAddr. A
+block of DevAddr addresses will be obtainable for an OUI via a blockchain
+transaction. This range of addresses will be assigned to devices joined to this
+OUI and hotspots will be able to do fast and simple 'range routing' to route
```
vihu: Can we explicitly define what exactly "range routing" is please. Maybe a glossary at the bottom of the doc would be useful?

0008-lorawan-join-routing.md
```diff
@@ -0,0 +1,294 @@
...
+much less ambiguous routing for large numbers of devices.
+
+XOR filters require inputs be hashed to a 64 bit integer before being stored or
+compared against the XOR filter. We propose the use of
+[xxhash](http://cyan4973.github.io/xxHash/)'s 64 bit mode. This is one of the
+fastest hash functions available and seems to provide good distribution for the
+XOR filter. Both the DevEUI and the AppEUI (totaling 128 bits together) would be
+hashed to a 64 bit key, this should help in cases of DevEUI collisions.
+
+A new ledger entity would be added to the blockchain, a 'routing entry'. These
```
evanmcc: The language in the next few paragraphs is unclear.  Is there one xorfilter per OUI or one big one?

Vagabond: At least one per OUI, for very large OUIs there might be several (so we limit the size of any single object we have to examine).

0008-lorawan-join-routing.md
```diff
@@ -0,0 +1,294 @@
...
+```
+
+The benchmark code can be found
+[here](https://github.com/Vagabond/bloom_router). These runs were generated by
+the command in the README. The benchmark was first run on the Intel machine to
+generate the routing tables, run a second time to use the generated routing
+tables, and then the tables were copied to the Pi to run the test with the same
+databases (generating the tables can be quite RAM intensive, so it is unsuitable
+to be done on the Pi).
+
```
evanmcc: I'd love to see some extension briefly discussing the process of generating the table or tables.  Who do we trust to do this?  How do we certify the validity of a table?  How do we keep a customer from shooting themselves in the foot?

0008-lorawan-join-routing.md
```diff
@@ -0,0 +1,301 @@
...
+| 1,000     | 2.48 kbytes    |
+| 10,000    | 24.09 kbytes   |
+| 100,000   | 240.21 kbytes  |
+| 1,000,000 | 2.34 mbytes    |
+
+With the largest XOR filter, we can check for key membership in about 14
+milliseconds on an raspberry pi 3b+ in 32 bit mode. The smallest xor filter can
+be checked in well under a millisecond.
+
+In a more complete test, 10,000 OUIs holding device space for 50 million devices
```
evanmcc: Are we doing anything clever here, or just checking them one by one?

0009-lorawan-devaddr-routing.md
```diff
@@ -0,0 +1,195 @@
...
+```
+running with 1000 OUIs and 8000000 Devices
+Attempting 1000 random lookups
+Average memory 0.08Mb, max 0.10Mb, min -0.06Mb
+Approximate database size 0Mb
+Average lookup 0.03s, max 0.4s, min 0.0001s
+Lookup misses 0
+```
+
+The good news is that, on average, this is orders of magnitude faster, on
```
vihu: nit, double "on average"

0008-lorawan-join-routing.md
Vagabond: Added something

0008-lorawan-join-routing.md
Vagabond: We have to do a linear scan, as far as I know.

PharkMillups:
This is mostly done. @Vagabond is going to give it one more review then we'll merge.

jamiew:
In the interest of keeping PRs moving, I am closing this as `stale`. Please re-open if you'd like to finalize and merge it for discussion.

PR #11, opened 2020-02-19
miner update HIP

fvasquez:
Rendered view: https://github.com/helium/HIP/blob/e4a3b70c8f6234c49b9264156ba0ea250d5ac705/0000-miner-update.md


0000-miner-update.md
```diff
@@ -0,0 +1,199 @@
+- Start Date: 2020-02-13
+- HIP PR: <!-- leave this empty -->
+- Tracking Issue: <!-- leave this empty -->
+
+# Summary
```
lthiery: I think this Summary exposes the problem and its scope well, but I think it would be helpful to have a summary of proposed solution.

0000-miner-update.md
```diff
@@ -0,0 +1,199 @@
...
+
+Another solution is to package miner releases in Docker containers for
+deployment to and execution on Hotspots. Instead of pointing their Hotspots at
+alternative minerhaul servers Hotspot owners would instead point them at
+alternative Docker registries. A miner package maintainer uses Docker push
+commands to publish new miner images to her registry. Hotspots use Docker pull
+commands to fetch new miner images from their designated registries. A Docker
+runtime and client-side scripts will need to be added to nextgate for automatic
+teardown and setup of miner images on the Hotspots.
+
```
lthiery: What would be the user interface to configure the source of updates? App BLE service?

0000-miner-update.md
```diff
@@ -0,0 +1,199 @@
...
+user deploying a homemade firmware update that accidentally bricks their
+Hotspots is simply too high. What we want is a way for Hotspot owners to only
+upgrade the miner portion of their devices' firmware so that their Hotspots
+remain online and continue to receive hardware-specific updates (Linux kernel,
+gateway-config, LoRa packet_forwarder) only from Helium.
+
+The simplest solution is to install miner and all its dependencies (RocksDB,
+libsodium, ebus) as an image overlay on top of a Hotspot's root file system.
+This extraction step would happen at boot time once Linux has started and the
+immutable root file system has been loaded into memory. The end result is
```
lthiery: So the user needs to push a whole tarball out to their hotspot somehow? If there was a UX to indicate the source of the update, we would need to make that service overwrite some SD card contents to store the tarball? or I suppose you could store it in `var`?

0000-miner-update.md
```diff
@@ -0,0 +1,199 @@
...
+- What parts of the design do you expect to resolve through the implementation
+of this feature?
+
+With the exception of image overlays all of the approaches being considered are
+new so the only way to assess them is through rapid prototyping.
+
+- What related issues do you consider out of scope for this HIP that could be
+addressed in the future independently of the solution that comes out of this HIP?
+
+How does a third party package and sign a miner release for download and
```
lthiery: perhaps once you go into third-party mode, no more signing?

0000-miner-update.md
```diff
@@ -0,0 +1,199 @@
...
+We are not abandoning the Buildroot embedded Linux build system for existing
+first generation of Helium Hotspots so any miner update approach we adopt will
+need to integrate with Buildroot.
+
+# Success Metrics
+[success-metrics]: #success-metrics
+
+- What should we measure to prove a performance increase or decrease?
+
+We need measure the impact on blockchain throughput that running miner inside a
```
lthiery: This was my primary concern with the docker approach, although it does feel more modern/flexible.

lthiery: Do we use any specialized crypto processor instructions in miner? Are those preserved via Dockerization?

0000-miner-update.md
```diff
@@ -0,0 +1,199 @@
...
+
+The final solution will need to include clear step-by-step instructions for
+miner package or image maintainers. This documentation will likely take the form
+of a GitHub project README for a Helium minerhaul repo that miner package
+maintainers will need to fork to deliver their updates.
+
+UI for entering the URL of an alternative miner package repository will need to
+be added to the Hotspot Settings page of the Helium mobile app. A related help
+article will also need to be added to the support section of Helium's website
+for end users.
```
fvasquez: From @cokes518 

> We've seen previously that if the miner is in a crash loop on the Hotspot, a user would not be able to pair with it over BLE and thus could never get to their Hotspot Settings to change or revert the URL of a non-helium repo.
> 
> There's a high likelihood of this happening. I'd suggest including the ability for a user perform a button-boot to revert to a working version of Helium miner.

0000-miner-update.md
```diff
@@ -0,0 +1,199 @@
...
+nextgate to download and apply miner updates.
+
+Another solution is to package miner releases in Docker containers for
+deployment to and execution on Hotspots. Instead of pointing their Hotspots at
+alternative minerhaul servers Hotspot owners would instead point them at
+alternative Docker registries. A miner package maintainer uses Docker push
+commands to publish new miner images to her registry. Hotspots use Docker pull
+commands to fetch new miner images from their designated registries. A Docker
+runtime and client-side scripts will need to be added to nextgate for automatic
+teardown and setup of miner images on the Hotspots.
```
fvasquez: From @Vagabond 

> Propose we proceed with Docker until we find a solid reason it's unsuitable.

fvasquez: That's a good point. A user can still roll back to a previously working version of the Hotspot firmware by button-booting. However, a better fix is for not being able to pair over BLE is for the GATT server to read a Hotspot's name from the ECC rather than ask miner for it.

0000-miner-update.md
fvasquez: That's a good point. A user can still roll back to a previously working version of the Hotspot firmware by button-booting. However, a better fix is for not being able to pair over BLE is for the GATT server to read a Hotspot's name from the ECC rather than ask miner for it.

fvasquez: From @Vagabond 

> Yeah, we should fix this irregardless so a working miner is not required for BLE to work.

0000-miner-update.md
fvasquez: We want to use ARM's NEON SIMD extension and ARMv8 crypto extensions. I believe we are already cross-compiling OpenSSL with the former but we are not yet cross-compiling RocksDB with the latter. Docker now supports multi-arch images including arm64 so I think those instructions are preserved.

0000-miner-update.md
fvasquez: > So the user needs to push a whole tarball out to their hotspot somehow?

Yes.  They would push their own miner updates via a minerhaul GitHub fork that they deploy to Heroku.

> We would need to make that service overwrite some SD card contents to store the tarball?

Correct.  We could do an atomic move like we do for the embedded Linux image updates in a `MINER_PACKAGE_DIR` inside `var`.

0000-miner-update.md
fvasquez: The only way to configure a Hotspot right now is through the mobile app so some write-only characteristics will need to be added to the BLE GATT server: MinerUpdateSourceUrl, SslPublicCert, SignifyPublicKey.

0000-miner-update.md
fvasquez: We could eliminate signature verification.  That increases the risk of a Hotspot being hijacked for a botnet though.

0000-miner-update.md
fvasquez: Sounds like a plan.

0000-miner-update.md
fvasquez: Working on it.  See [helium/gateway-config#54](https://github.com/helium/gateway-config/pull/54).

0000-miner-update.md
fvasquez: Done.

Vagabond:
@mrallen1 and I have been playing around with Open Container Initiative (OCI) images. I think we can mount and boot them on the hotspot (using archivemount and chroot) without installing any container runtimes. This would, however, allow other people to run the container in docker/oci/etc on non hotspots as well.

Vagabond:
@mrallen1 and I have been playing around with Open Container Initiative (OCI) images. I think we can mount and boot them on the hotspot (using archivemount and chroot) without installing any container runtimes. This would, however, allow other people to run the container in docker/oci/etc on non hotspots as well.

I also think we could boot a OCI image using https://github.com/containers/bubblewrap
```
(base) ^andrew@system76:: ls -lah `which bwrap`
-rwxr-xr-x 1 root root 55K Nov 30 11:01 /bin/bwrap
(base) ^andrew@system76:: ldd `which bwrap`
        linux-vdso.so.1 (0x00007ffc36bf9000)
        libcap.so.2 => /usr/lib/libcap.so.2 (0x00007f833823d000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f833807a000)
        /lib/ld-linux-x86-64.so.2 (0x00007f8338279000)
```

```
(base) ^andrew@system76:: ls -lah `which docker`
-rwxr-xr-x 1 root root 81M Feb 16 21:10 /bin/docker
(base) ^andrew@system76:: ldd `which docker`
        linux-vdso.so.1 (0x00007ffd58be2000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f5ad85dd000)
        libdl.so.2 => /usr/lib/libdl.so.2 (0x00007f5ad85d8000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f5ad8415000)
        /lib/ld-linux-x86-64.so.2 (0x00007f5adcbf1000)
```

Vagabond:
More concretely I propose we'd make the following miner containers:

- amd64
- arm64 full
- arm64 slim (does not include Erlang's runtime)
- arm32 full

We'd run the arm64 slim container on the hotspot, it would essentially just contain /opt/miner and we'd mount through the needed system paths into the bubblewrap environment. The other containers would contain a full system image and an Erlang distribution (these could be done via composing layers). They would be served through a 'container registry' such that 'docker pull' or suchlike would be compatible. Most cloud platforms support a compatible cloud registry API now (eg Amazon ECR) and open source implementations exist. These registries can already contain OCI images.

https://docs.docker.com/registry/spec/api/
https://docs.microsoft.com/en-us/azure/container-registry/container-registry-oci-artifacts
https://github.com/opencontainers/artifacts

Vagabond:
The v2 registry can be queried with bash/jq via something like https://github.com/gforghetti/docker_registry_api_bash_functions

mrallen1:
I made [a story in Clubhouse](https://app.clubhouse.io/hlm/story/6497/containerize-miner-in-oci-format) for this work.

fvasquez:
> I made a story in Clubhouse for this work.

@Vagabond @mrallen1 Your proposed OCI and bubblewrap solution is interesting. I do have my doubts as to how mature the bubblewrap tooling is compared to Docker's. Namely does bubblewrap really have something akin to `docker pull` and can it authenticate image registries and verify signed images. Regardless of these concerns I find a chroot-based solution like this one compelling enough to warrant further investigation. I will incorporate an OCI/bubblewrap option into this HIP.

Vagabond:
> > I made a story in Clubhouse for this work.
> 
> @Vagabond @mrallen1 Your proposed OCI and bubblewrap solution is interesting. I do have my doubts as to how mature the bubblewrap tooling is compared to Docker's. Namely does bubblewrap really have something akin to `docker pull` and can it authenticate image registries and verify signed images. Regardless of these concerns I find a chroot-based solution like this one compelling enough to warrant further investigation. I will incorporate an OCI/bubblewrap option into this HIP.

Bubblewrap relies on cgroups, same as Docker does (it's just a much thinner shim over the top). Bubblewrap is used in a ton of places, including flatpak applications and I found out yesterday it's how void linux sandboxes their package building.

https://wiki.archlinux.org/index.php/Bubblewrap
https://github.com/void-linux/void-packages#requirements

It definitely does *not* have a docker pull implementation, but the docker protocol is HTTP based, and as I linked above, shell libraries exist to query it (so we'd do docker pull via bash/curl).

I've been thinking about the OCI "layers" we might want to use:

* Base image (eg alpine-linux with Erlang)
* Portable BEAM files and release scripts/configs
* NIFs and other native code

This would allow us to reuse 'layer 2' for all of our docker images and just crosscompile layer3 for arm/arm64 (and swap out the base image for an architecture appropriate image). As a bonus, since our NIFs don't change a lot, we could simply update the middle layer the majority of the time.

Vagabond:
Additionally we can do stuff like @macpie did in the router docker stuff, where we use the environment to configure a sys.config.src template at runtime (for things like paths, key mode, etc)

PR #12, opened 2020-02-19
sign miner HIP

fvasquez:
Rendered view: https://github.com/helium/HIP/blob/0484b9fac67b3590bcbd3cafc8228d363fc0112b/0000-sign-miner.md


0000-sign-miner.md
```diff
@@ -0,0 +1,159 @@
...
+
+# Summary
+[summary]: #summary
+
+This is a proposal to separate Helium's blockchain software from the rest of the
+Hotspot firmware image. Decoupling Helium's blockchain software from our
+consumer Hotspot hardware and mandatory over-the-air updates makes a stronger
+case for decentralized nature of the Helium Network. While Helium Systems Inc
+makes Hotspots it may not necessarily be the same entity that owns the rights to
+the Helium blockchain software.
```
fvasquez: From @Vagabond 

> I would like to see this section fleshed out more, if possible.

0000-sign-miner.md
```diff
@@ -0,0 +1,159 @@
...
+
+# Success Metrics
+[success-metrics]: #success-metrics
+
+Success is measured by:
+
+- time it takes to Helium Blockchain and Embedded teams to implement a signing
+and verification solution
+
+- effort required by Helium Blockchain team to issue a miner release
```
fvasquez: From @amirhaleem 

> Is there a way that this could be architected such that the miner is an entirely separate OTA function than nextgate? So on a Helium Hotspot nextgate would be a mandatory update, but a user could opt to change where miner comes from and deal with the appropriate keys/signature verification for that repo (similar to adding a custom apt repo)? As a Hotspot owner I'd be able to configure (via BLE) which OTA server/keys I want to use for the miner part of the firmware.
> 
> My understanding of what you've written above is that a user still has no way of, for example, specifying their own miner release with modified code if they choose to.

fvasquez:
> In coordination with [the other HIP](https://github.com/helium/HIP/pull/11), I think I'm not understanding the intent here.
> 
> I thought decoupling miner signing from full firmware image signing was the goal, but this GPG approach seems to only provide verification at the miner build step? Meanwhile, the intent of the miner update HIP was to enable a distinct miner update from the firmware image update process. Therefore, verification needs to occur at the update not at the build step?

@lthiery this HIP was written to address the following topic of discussion in isolation.

2) How to think about separating consensus/blockchain related firmware signing from Hotspot-specific bits such that a separate entity (like a foundation) could hold signing power for the blockchain part

I purposely wrote it before [the other HIP](https://github.com/helium/HIP/pull/11) so that topic (see 3) would not overly-constrain the design space and prematurely influence my decisions.  

3) How to allow Hotspot owners to specify and deliver their own firmware updates

I wanted to capture what the easiest ways to sign miner software releases are in the absence of alternative OTA firmware updates.  After talking to @amirhaleem I now understand that whole point of separating miner for signing is to enable alternative OTA firmware updates. 

You are correct in concluding that signing a Git tag (this GPG approach) or signing a gzipped tarball of the source (roughly equivalent) at some Git tag does not work for OTA firmware updates.

You'll also notice that the scope of the alternative OTA firmware updates was reduced to just the miner. The idea to update miner independently of the rest of the Hotspot firmware arose from discussions with the Blockchain Engineering team.

0000-sign-miner.md
```diff
@@ -0,0 +1,159 @@
...
+tarball release can be cryptographically signed so that a package manager or
+build tool can verify that the code is blessed by a particular signatory.
+
+The Git version control system has built-in support for signing and verifying
+tags using GPG encryption. Note that signing a tag is different from signing the
+source tarball or gzipped binaries for a tag. A Git tag is just an alias to a
+commit hash. A package manager or build tool uses Git to verify a release tag
+against the GPG public key of the expected signatory. If verification succeeds
+then the build tool clones the contents of the Git repo at that tag and builds
+the release binaries for deployment.
```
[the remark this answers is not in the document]
fvasquez: It doesn't unless you consider the contents of a Git repo at a given commit hash to be an artifact.

0000-sign-miner.md
fvasquez: Done.

0000-sign-miner.md
fvasquez: That's correct. How to a point a miner at homemade miner release is addressed by PR #11.

PR #16, opened 2020-02-26
How Data Credits are Expensed

lthiery:
Initial proposal after in-person discussion.

Rendered view: https://github.com/helium/HIP/blob/83df9ee5013f5d9224494f63ab433d6baf742eff/0002-DataCredits.md


0002-DataCredits.md
```diff
@@ -0,0 +1,160 @@
...
+[detailed-explanation]: #detailed-explanation
+
+When an Operator transmits or receives a packet, they are unable to transmit 
+and/or receive packets. As such, we must fight against exploits and charge a fee
+for Join and Join-Accept frames.
+
+When a User considers the Helium Network, they know about their application, and
+so have ideas of how large their payloads are, but may not know of the nuances
+of LoRa or LoRaWAN versus other modulations or protocols. To help them make the
+business decision to use the Helium Network, we believe it to simply the data cost
```
amirhaleem: the grammar in this sentence is a little odd

lthiery: Agreed. I did some rephrasing

Vagabond:
I think this is generally correct. I would add, however, that there's additional overhead in the network exchange between the hotspot and the router that means that there won't be a 1:1 correspondance between bytes on the air and bytes on the wire. If you're operating a hotspot with expensive backhaul this might be an unpleasant surprise.

PR #22, opened 2020-06-04
A process for managing Helium Improvement Proposals

jamiew:
Defines an initial process for proposing, discussing, and finalizing improvements to the Helium blockchain using Helium Improvement Proposals (HIPs)

Rendered view: https://github.com/helium/HIP/blob/a2e5561c9cacdd93c970f99029947895693d5aac/0007-managing-hip-process.md

There are a number of open questions enumerated at the end, and the "Discussion & approval" process in particular could use more detail. As well as e.g. starting the community calls that this document mentions.

### DIFF
--- a/0007-managing-hip-process.md
+++ b/0007-managing-hip-process.md
@@ -1,0 +1,116 @@
+
+# Managing Helium Improvement Proposals (HIP000)
+
+- Start Date: 2020/05/28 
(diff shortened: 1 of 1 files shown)


PharkMillups:
Merging this in. I think we're going to end up pulling most of this into the README but this is a good place for this for the moment.

PR #30, opened 2020-06-19
HIP(x): Remote Location Assert

rawrmaan:
Since the GPS location assertion check will soon be enforced, more than 1500 hotspots will need to be adjusted in some way in order to have a good GPS fix and continue participating in Proof of Coverage (see current hotspot `gps_fix_quality` states below). For many hotspots, this will require updating the asserted location to accurately reflect the hotspot's latest location, which can currently only be done with physical access to the hotspot.

```
gps_fix_quality | count 
----------------+-------
good_fix        |  2275
no_fix          |  1090
bad_assert      |   514
```

This HIP proposes a new transaction, `assert_location_v2`, which will allow a hotspot's location to be asserted remotely without physical access to the hotspot.

### DIFF
--- a/remote-location-assert.md
+++ b/remote-location-assert.md
@@ -1,0 +1,88 @@
+- Start Date: 2020-06-18
+- HIP PR: <!-- leave this empty -->
+- Tracking Issue: <!-- leave this empty -->
+
+# Summary
+[summary]: #summary
+
+Since the GPS location assertion check will soon be enforced, more than 1500 hotspots will need to be adjusted in some way in order to have a good GPS fix and continue participating in Proof of Coverage (see current hotspot `gps_fix_quality` states below). For many hotspots, this will require updating the asserted location to accurately reflect the hotspot's latest location, which can currently only be done with physical access to the hotspot.
+
+```
+gps_fix_quality | count 
+----------------+-------
+good_fix        |  2275
+no_fix          |  1090
+bad_assert      |   514
+```
+
+This HIP proposes a new transaction, `assert_location_v2`, which will allow a hotspot's location to be asserted remotely without physical access to the hotspot.
+
+# Motivation
+[motivation]: #motivation
+
+Asserting a new location with the `assert_location_v1` currently requires a signature from the hotspot, which requires physical access (Bluetooth connection) to obtain.
+
+This poses a massive operational headache for owners and Patrons who have deployed hotspots at remote locations, and leaves them with no options other than:
+
+- a) travel to the hotspot's current location
+- b) have the hotspot temporarily shipped back
+- c) share their private key with hotspot hosts to perform the assertion
+
+All of these options are costly and/or dangerous, and options a & b prevent iteration on the hotspot's location to get it "just right" within the GPS fix tolerance range. Iteration is important because a consistent GPS fix can take hours or days to measure.
+
+# Stakeholders
+[stakeholders]: #stakeholders
+
+Any Helium hotspot owner who has deployed at least one remote hotspot. The issue of not being able to assert location remotely has caused a large outcry within the hotspot owner community on Slack and Telegram.
+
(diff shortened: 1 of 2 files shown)


jamiew:
As a HIP this looks good to me and I'd vote to merge for discussion (and code)

abhay:
Can we merge and continue discussion, @PharkMillups @vihu ?

remote-location-assert.md
```diff
@@ -0,0 +1,88 @@
...
+Any Helium hotspot owner who has deployed at least one remote hotspot. The issue of not being able to assert location remotely has caused a large outcry within the hotspot owner community on Slack and Telegram.
+
+The Helium blockchain engineering team.
+
+# Detailed Explanation
+[detailed-explanation]: #detailed-explanation
+
+## Implement a new transaction, `assert_location_v2`, and optionally deprecate the previous `assert_location_v1`
+
+This new transaction would require the owner's private key, hotspot public key and lat/long coords of the new desired location. It would be submitted to the blockchain to replace the hotspot's existing location, and charge the same assert fee (in DC) as `assert_location_v1`.
```
vihu, line 45: Just so we are clear on this, the existing `assert_location_v1` txn does _not_ have lat/long coords of the location, instead it contains an `h3 index` (refer: https://github.com/uber/h3) which encapsulates the lat/long information at h3 resolution 12.

Furthermore, we may add additional fields to `assert_location_v2` txn such as the `OUI` a hotspot may belong to, just so we have that data on chain while we rev the assert location txn. Refer: https://github.com/helium/proto/pull/41/files (WIP)

remote-location-assert.md
```diff
@@ -0,0 +1,88 @@
...
+Not making this change could deter current and future owners from making remote deployments, which could be viewed as too much of an operational hassle. This could hurt the growth of the network.
+
+Hotspot owners could be upset if they aren't able to update a far-away hotspot before the GPS assertion check is enforced and they lose their earnings until they can access it.
+
+It will generally reduce operational friction, and friction in the community.
+
+# Unresolved Questions
+[unresolved]: #unresolved-questions
+
+Are there any security, privacy or other implications of allowing location to be remotely asserted with only the account holder's private key and not the hotspot's signature?
```
vihu, line 74: Not having the hotspot's signature in `assert_location_v2` transaction should be fine.

remote-location-assert.md
abhay: I say land this PR and @vihu would you be willing to update the HIP to include this?

vihu: @abhay I believe @rawrmaan might need to update the wording a bit since this is on his fork of the hip repo :) I don't have any concern landing this once that's done.

rawrmaan: Done!

jamiew:
@rawrmaan I was too slow to merge this and now we have a HIP10 and draft HIP11 and HIP12. How insulted would you be if I number this HIP13, despite predating all 3 of those? Could also number this 11 and renumber 12 and 13

rawrmaan:
@jamiew How will my friends and family respect me if this can't be HIP10, 11, or 12? This is devastating.

I suppose I'll have to settle for 13 and regain my honor through other feats.

abhay:
Haha. @rawrmaan. Also should rescope this for large scale deployers of gateways and establishing a way for them to manage their hotspots since the GPS fix data is not going to be used. 

I think it's still a worthwhile HIP that should be accepted for discussion.

rawrmaan:
@abhay How would you recommend we rescope this? I think the feature is still relevant as it is laid out. GPS may not matter, but hosts need to place their hotspots in the correct spots on the map if they move, or they will not be able to PoC with nearby hotspots.

jamiew:
Just kidding, this is actually HIP12

PR #37, opened 2020-08-20
Correct math for reduced monthly allocation

abhay:
Also clean up formatting.


Carniverous19:
looks correct to me,

abhay:
@amirhaleem @PharkMillups @Vagabond @refugeesus 👋

jamiew:
Merged. Wdyt of prefixing commits with the relevant HIP# in future? Don’t want to be overly pedantic but might make commit history easier to process quickly

PR #46, opened 2020-10-01
Updates to HIP7

anthonyra:
Try number 3...

### DIFF
--- a/0007-managing-hip-process.md
+++ b/0007-managing-hip-process.md
@@ -2,8 +2,12 @@
 
 - Start Date: 2020/05/28 
 - Original HIP PR: [#22](https://github.com/helium/HIP/pull/22)
-- Tracking Issue: [#26](https://github.com/helium/HIP/issues/26)
-- Authors: [@jamiew](https://github.com/jamiew)
+- Tracking Issue: Pull Request (PR) comments <- This will be left blank for author submit of PR, HIP Editor will specify location ->
+- Authors: [@jamiew](https://github.com/jamiew), [@anthonrya](https://github.com/anthonyra)
+
+- Category: Meta / Process
+- Premission Level: Idea
+- Status: Draft
 
 
 # Summary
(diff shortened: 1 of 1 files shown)


0007-managing-hip-process.md
```diff
@@ -21,74 +25,136 @@ As Helium Inc begins to share control over the development and management of the
...
+We do not aim to adjudicate routine technical changes to miner software or the day-to-day dealings of Helium Inc. We expect that any change that could have significant impact on the functioning of the Helium Network would be vetted through the HIP process. This will be discussed in more detail below in the categories of HIP.
+
 
 # Stakeholders
 [stakeholders]: #stakeholders
 
-We plan to circulate this proposal in [Slack #helium-improvement-proposals](https://chat.helium.com) and on the [Helium community forum](https://community.helium.com). We will also solicit specific individuals for feedback, from both within the Helium community and other crypto and open-source projects. 
+We plan to circulate this proposal in [Discord #hips](https://discord.gg/helium) and on the [Helium community forum](https://community.helium.com). We will also solicit specific individuals for feedback, from both within the Helium community and other crypto and open-source projects.
```
abhay, line 34:
```suggestion
We plan to circulate this proposal in [Discord #hips](https://discord.gg/helium). We will also solicit specific individuals for feedback, from both within the Helium community and other crypto and open-source projects. 
```

abhay (commented):
Some suggestions.

jamiew:
Thanks for all the suggestions here. I have questions about a few ideas:

* Is the motivation behind the `Permission Level` of HIPs largely to justify "emergency" actions like HIP10, or is it for something else? If so, would it be possible to simplify it down to a proposal to just how we handle emergency actions like that?
* Do you have an example of a proposal that falls into the `Blockchain Performance` category that wouldn't be addressed by `Technical`?

anthonyra:
> Thanks for all the suggestions here. I have questions about a few ideas:
> 
> * Is the motivation behind the `Permission Level` of HIPs largely to justify "emergency" actions like HIP10, or is it for something else? If so, would it be possible to simplify it down to a proposal to just how we handle emergency actions like that?

Yes, that way anybody looking at the HIP knows what kind of timeline to expect. It’s an attempt to start tying in Governance Control with HIP. Now I also think it’s important to note the difference between an `Urgent` and bug. `Urgent` is when the source code is functioning as intended but the results are detrimental to the blockchain. `Bug` is where there’s an issue with the source code that comprises its ability to do what was originally intended.

> * Do you have an example of a proposal that falls into the `Blockchain Performance` category that wouldn't be addressed by `Technical`?

Pretty much all Blockchain Performance will fall under technical I just suggested the complete separation so that when a user writes up a HIP about Blockchain Performance it locks them into the more condensed timeline. When you look at an operating procedure the less if thens you can have the better so that an issue doesn’t get miss labeled then wrongfully handled. It’s opinion based but I’ve seen issues arise when you rely on the person to read the FI statement and make the right choice to proceed.

jamiew:
Tentatively I'd prefer to focus the changes just on addressing the "Emergency Action" idea rather than adding the whole new "Permission Level" concept, and same for keeping categories to a minimum. e.g. right now we _only_ have Technical proposals, so I'm not sure if we even need categories at all.

I'll give this another read since there are a number of other fixes I'd like to merge, but just want to do a copy & grammar pass too

Thanks again for your contribution

anthonyra:
No worries, I can understand the importance of simplicity just old habits die hard for me wanting to be more specific and Black and White with things.

jamiew:
Gonna close this one for now to keep repo moving, but happy to incporate one that just focuses on the "emergency vs. normal schedule" concept

PR #48, opened 2020-10-01
Update HIP12: Remove references to GPS fix

rawrmaan:
Removes reference to GPS fix, which no longer exists in the system. This feature remains useful without GPS fix requirement.


abhay:
LGTM. Thanks, @rawrmaan. Definitely would love to see the community come to a consensus on the demand for this update.

PR #53, opened 2020-10-12
Beacon rewards

Carniverous19:
See commit message, added some example topology comparisons between multihop and beaconing rewards


jamiew:
LGTM, squashin' and mergin'

PR #56, opened 2020-10-15
Proposed changes to HIP16 "random-consensus-group-election.md"

tjain-mcc:
Rendered view of updated version: https://github.com/helium/HIP/blob/af60c836913cba2fc56ee4e1202921dcd5c1174f/0016-random-consensus-group-election.md

Diff: https://github.com/helium/HIP/pull/56/files

### DIFF
--- a/0016-random-consensus-group-election.md
+++ b/0016-random-consensus-group-election.md
@@ -1,10 +1,10 @@
-# HIP
+# HIP16: Random Consensus Group (CG) Election
 
-- Author(s): realPaulM
+- Author(s): [@PaulVMo](https://github.com/PaulVMo) (@PaulM on Discord)
 - Start Date: 2020-10-14
 - Category: Economic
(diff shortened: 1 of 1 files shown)


jamiew:
Tagging in @PaulVMo as original author -- feel free to comment on these changes and bless or reject a merge

evanmcc:
My vote would be to table this until we (Helium) can put together a comprehensive proposal that involves staking.  I think that adding staking to score removal will needlessly delay score removal as we work out the political aspects of staking and the further evolution of the consensus group.

PaulVMo:
I tend to agree with the idea of keeping staking as a separate though equally important HIP. There is no reason that can't be a HIP right now as well. How can we get broader input on whether or not to do an interim step of removing score?

jamiew:
@PaulVMo @tjain-mcc I'm understanding that this means this specific patch is rejected, and it's suggested to move staking-related concepts to their own HIP. 

Please re-open if I got that wrong!

PR #61, opened 2020-10-21
Update HIP 16 to remove notion of a "CG score"

abhay:
Since we're using a random function to elect/deselect, we shouldn't bother calling it a score.

cc: @jamiew @PaulVMo


jamiew:
Merging per @PaulVMo comments in Discord

PR #66, opened 2020-10-29
Governance for third party gateways (HIP draft) 

philltran:
Starting a draft of a HIP so the community can flesh out the governance for third party gateways.

Rendered view: https://github.com/philltran/HIP/tree/governance-for-third-party-gateways/00xx-governance-for-third-party-gateways.md


jamiew:
See also #15 for existing ideas on how to trust DIY hotspots

philltran:
@jamiew Thanks for that reference to HIP 9. My thought is that this HIP is more focused on establishing the criteria to trust the _**vendor**_ that is making the gateways / hotspots. By proxy we will need to include ideas from HIP9, but I don't want this to get hung up on trying to solve the "DIY hotspot" problem. I think the solution should still happen in HIP 9. Let me know if you think otherwise.

philltran:
@jamiew  With PR #86 should I close this PR?

jamiew:
@philltran yeah I think that makes sense 👍

PR #67, opened 2020-10-29
Update approved HIPs

PaulVMo:
Updates to status for recently approved HIPs per comments on the tracking issues; added definitions for the different statuses


jamiew:
Approved approved approved 🙏

PR #68, opened 2020-10-29
Golden Tabs (HIP draft)

lthiery:
Rendered: https://github.com/helium/HIP/blob/lthiery/golden-tab/0020-golden-tabs.md

### DIFF
--- a/0020-golden-tab.md
+++ b/0020-golden-tab.md
@@ -1,0 +1,148 @@
+# HIP18: Remove Oracle Forecast for DC Burn
+
+- Author: [@lthiery](https://github.com/lthiery)
+- Start Date: 2020-10-29
+- Category: Technical
+- Original HIP PR: 
+- Tracking Issue: 
+
+# Problem Statement
+[probem-statement]: #problem-statement
+In its current state, the Helium Blockchain does not allow permissionless adding of gateways. All gateways on the 
+blockchain have one of the following origins:
+
+|Name|MSRP|Quantity|Shipping Dates|Keys Store|
+|---|---|---|---|---|
+|Genesis Hotspots|$0|45|2019-07-29|File|
+|Helium Hospots|$500|~9k on-chain, 14.5k in staking server|2019-07-29-2020-10-01|ECC608|
+|RAK Hotspots|$250|   |2020-10|File|
+|Miner Pro Alpha|$0|339|2020-09|File|
+
+
+This strategy has thus far prevented large-scale attacks of “virtual gateways” as permissionless adding has the proposed
+ cost of $40 to add and $10 to assert location. Instead, attacks require buying hardware listed above with their 
+ respective costs; this means being a good actor is slightly easier as you are guaranteed to have LoRaWAN hardware and
+ the cost of each node is at least 4x the cost of the permissionless model.
+
+It is paramount to the network's growth in coverage to allow the permissionless adding of gateways to the network. In 
+ other words, anybody must be able to source a standard LoRaWAN gateway (or use already deployed hardware) and pay the
+ $40 staking fee.
+
+It is a general consensus in the community, including Helium and DeWi, that Proof-of-Coverage (POC) does not yet do 
+ enough to instill confidence for permissionless adding of gateways to be enabled. There are many ideas for how to 
+ improve POC, but they all boil down to concepts of circular trust validation between gateways. 
+
+HIP-9 proposes a levelling scheme based on trustworthiness of the gateways based on origin; however, it is the opinion
+ of this HIP that gateways, no matter the origin, carry no inherent trust. In particular, it is quite possible to 
+ “jailbreak” even a Helium Hotspot and to run an arbitrary Miner on the hardware. And over the months we have seen many 
+ question deployments of Helium Hotspot clusters which have proven to be quite difficult to snuff out using circular 
+ validation logic.
+ 
+
+# Solution Summary
+[solution-summary]: #solution-summary
+
+This proposal side-steps the inherent lack of trust in gateways by creating a new blockchain actor: a “Golden Tab”. In 
+ effect, this actor creates a root of trust for levelling and POC.
+
+The Golden Tab (GT) is a tamper-proof GPS tracker, designed by Helium and audited by DeWi. The design is such that 
+ physical tampering may be near impossible and many public and confidential methods may be used to counteract GPS 
+ spoofing. The design also focuses on making the device secure and cheap, such that supply of Golden Tabs can be
+ relatively unrestricted.
+
+The GT pairs with a smartphone over BLE, upon which a p2p link with other network actors may be extended without adding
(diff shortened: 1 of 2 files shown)


mfalkvidd:
Some feedback. Feel free to discuss here or on Discord or wherever suitable.

## Untapped potential
I think the gravity of the following statement can be lost to many people
> It is paramount to the network's growth in coverage to allow the permissionless adding of gateways to the network. In other words, anybody must be able to source a standard LoRaWAN gateway (or use already deployed hardware) and pay the $40 staking fee.

Semtech [claims](https://www.semtech.com/lora) there are more than 1M LoRa-based gateways deployed world-wide. This means the current helium-compatible gateway count is less than 2% of the total gateway count, leaving 98% untapped potential.

Helium has quickly become a large player in the LoRaWAN ecosystem, and has great momentum. But most people don't realize that compared to the global pool of gateways, Helium is still pretty small. Helium could grow faster if third party gateways were allowed on the chain.

Sugegstion: Add something along the lines of the above text to the problem statement section, to clearly quantify the untapped potential.

## Incentive
For hotspot owners, there is an incentive to buy the Golden Tabs, and to revisit the area covered by their gateways as often as required, because otherwize their earnings will be hurt. But hotspot owners might not be best equipped to revisit. Imagine a hotspot provider with hotspots in 30 cities. How would they revisit each city efficiently?

Maybe an incentive could be created for non-hotspot owners? This could be a bit similar to the model used by some scooter companies to get people to bring in scooters for charging ([reference](https://www.theverge.com/2019/3/15/18267128/lime-electric-scooter-charging-juicers-harvesting-business)). The scooter companies have been accused for not being environmentally sustainable though (driving around in old cars creates pollution) so we might have to be careful to avoid negative environmental impact, as well as negative publicity/PR.

## Privacy
Helium would be able to track the location of a Golden Tab whenever the Golden Tab transmits within coverage for a Helium gateway. This could be a privacy issue. Easiest way to handle this is probably to inform and get consent from anyone who carries an Golden Tab.

## Rural areas
Rural areas (such as those targeted by forest wildfire detection and prevention systems or agriculture use-cases) will need to be handled. The HIP suggests
> Therefore, high earning Gateways will require frequent validation, while perhaps remote and sparse clusters could subsist without frequent visits by a Golden Tab.

We'd need to find a balance here. For some areas, revisits more often than once per year might be impractical, even though there is quite a lot of DC traffic.

## Idea: fixed installation
Maybe the easiest solution to the revisit problem is to just leave a Golden Tab at each gateway location. Caveats:
* The placement would need a fairly unobstructed view of the sky to get GPS fix. Close to the hotspot antenna should work, since the hotspot antenna needs unobstructed view for good coverage.
* Battery life would have to be several years, to avoid frequent re-visits to charge the Golden Tab. Alternatively the Golden Tab could me mains powered, but getting mains power to a place where the Golden Tab has access to GPS signals might be tricky.
* The Golden Tab would probably be very close to the hotspot (due to cramped space, or challenge to protect the Golden Tab from theft and vandalism). This would mean that the Golden Tab needs to transmit at a low power, perhaps -10dBm to not overpower the receiver on the gateway.
* The BLE link to a smartphone would not be available
* Cost of the Golden Tab needs to be very low
* Production capacity of the Golden Tab needs to be high (potentially much higher than the production capcity of hotspots that are Helium-compatible, so we can utilize the untapped potential).

## Idea: Replace challengee with Golden Tabs
A big problem with third party gateways is that we cannot trust the asserted geographical position of the gateway (or that the gateway has coverage in that area). In a way, this HIP suggests to supplement the PoC process. How about instead of supplementing the PoC process, we let the Golden Tabs replace the challengee in the PoC process? That way we would still have strong incentives (or at least as strong as today). I realize that doing this would be quite controversial though, since PoC is very central to the Helium network.

## De-anonymizing Golden Tabs
The current HIP text:
> When a gateway receives a packet from a GT, it should not know of its origin. In other words, the packet looks like a standard LoRaWAN packet and will be routed to an OUI. To allow this, the GT construction should derive several mock LoRaWAN Sessions and the GT would maintain several LoRaWAN sessions at once such that OUI censorship of gateways is mitigated

It would probably be fairly easy for someone owning multiple hotspots to correlate received messages with when hotspots got a bump in score because the hotspot was re-validated, thus de-anonymizing the Golden Tabs. I agree that the Golden Tabs should look just like a normal LoRaWAN device. Multiple LoRaWAN sessions, as suggested in the HIP, will make de-anonymization harder but will it be hard enough?

I hope these comments can help discussion around this HIP.

lthiery:
@mfalkvidd I really appreciate the long-form feedback. We had some very interesting discussions on Discord yesterday, but I'm afraid the free-form back and forth is not only harder to digest for those trying to "catch up" on the conversation, but also will get lost historically.

Anyway, to respond to your feedback:

> Untapped potential

I think you are absolutely right to emphasize this and I've taken those Semtech marketing numbers into account. I might look at how to frame this more as well.

> Incentive

In general, my proposal is very thin on incentives other than "qualifying gateways for POC". I do think Golden Tabs need some kind of incentive mechanism for themselves.

While the scooter idea is an interesting one, I was imagining more of how an taxi or Uber drive could keep a Golden Tab on their dashboard for example and perhaps earn extra income. Thus their engagement with the network has nothing to do with being a coverage operator and being a coverage validator needs to be compensated in itself.

That being said, there are some thoughts about requiring staking/burning mechanisms for Golden Tab operators, which may create barriers than non-coverage providers may not be able to afford or risk. In effect, it creates a barrier to Golden Tabs, but also ensures some level of investment in the network.

Anyway, those are perhaps two things that need to be balanced, but it's clear that "coverage validation" is work and requires incentivization in itself.

> Privacy

Coverage validators are generating on-chain mapping data more or less and sharing it publicly in real-time, so yes, this should be made amply clear.

> Rural Areas

Some magical variables that configures time decay and decay relative to earnings, I think, will create the desired effect.

> Idea: fixed installation

Fixed installations aren't greatly compatible with the model I was imagining with the BLE connection, but it still seems within reach, technically (ie: dedicated device for the BLE -> p2p link). 

Your thoughts here though start having some overlap with the discussion we were having in Discord about potentially pivoting this idea into more of a Golden Gateway situation. I would still consider a requirement that the Golden Gateway model support mobile coverage mapping and, in effect, we could keep things almost the same as the current through-process here, but instead feature a SX130x with GPS timestamping of packets rather than a SX126x.

What troubles me is that the cost of Golden Gateways is going to be much higher, while I was thinking Golden Tabs could really be low barrier to entry.

> Idea: Replace challengee with Golden Tabs

I'm not opposed to this, but I think we could setup a Golden Tab solution in parallel and deprecate the old POC process if we deem the information/mechanism by POC has been supplanted by GT Challenges.

> De-anonymizing Golden Tabs

I did not make this clear enough, but my thought was that the LoRaWAN sessions would be re-derived every epoch, roughly, which might be when a score adjustment takes place. This way the GT can remain unknown to gateways.

lthiery:
I'm closing this for now in favor of a Golden/Anchor Hotspot concept.

PR #70, opened 2020-10-31
Discord community reference - Update 0007-managing-hip-process.md

cvolkernick:
Added in discord #hips reference for circulation.


jamiew:
Ah thank you. Will do another commit on top of this to remove Slack and Forum entirely, since they are no longer in use

Issue #65, opened 2020-10-29
HIP18: Remove Oracle Forecast for DC Burn


jamiew:
**Author(s):** @lthiery 
**Initial PR:** #62
**Start Date:** 2020-10-28
**Category:** Technical

**Rendered view:**
https://github.com/helium/HIP/blob/master/0018-remove-oracle-forecast-for-dc-burn.md

**Summary:**

A new oracle price takes effect one hour in the future, to provide users with predictability when converting HNT to DC. Unfortunately, this delay provides an on-chain arbitrage opportunity via state channels. Given the delay, an actor with HNT in-hand may see the forecasted drop in HNT price and thus convert HNT into DCs. After the drop takes effect, the actor may then “spend” the DCs in a state channel, naming a colluding gateway of theirs as the beneficiary. Effectively, the actor is burning and minting HNT on-chain at a profit with very little risk. This HIP seeks to address this flaw.

PaulVMo:
I like this. It is an elegant solutions to the arbitration opportunity.

One suggest for @lthiery - add some more detail about what "dynamic oracle price" means in the Solution section and how it would be implemented. I am assuming this means that a new oracle price (specifically for token burn only) would be calculated immediate upon receiving an oracle update rather than delaying 60 minutes.

M74774:
Did this HIP get consensus / voted on yet, and passed I hope?

jamiew:
Rough consensus has been achieved on this proposal and I'll be marking it as approved. On behalf of the community I'm also requesting the core development team begin any necessary engineering work.

This proposal has received nearly unanimous support from the community, and is debatably a "bug fix" that did not require a HIP at all. I applaud @lthiery and Helium Inc for submitting it through this process anyway, since it has been valuable to discuss and improve it in the open.

In informal Discord polling, 30/31 voters (97%) voted yes to approving this. The one `no` vote was from a user identifying themselves as Dapper Hickory Meerkat, the exploiter of this arbitrage hole 🙃: https://discord.com/channels/404106811252408320/773371214730625024/773944042295394304

Rough timeline:

* 2020/August -- `dapper-hickory-meerkat` is observed pushing a huge number of data credits through the system for unknown reasons
* 2020/10/12 -- the initial "get 150% of what you put in" data credits bug is fixed: https://engineering.helium.com/2020/10/12/blockchain-datacredit-reward-and-joinaccept-fixes.html
[...]

PR #77, opened 2020-11-11
HIP18: mark as Approved

PaulVMo:
Per @jamiew on Discord, it seems that rough consensus has been reached on HIP18 meaning it is approved


jamiew:
@PaulVMo since we're this close to community call I'll just hold off till then to formally update this. Thanks for the PR

jamiew:
Related to #65

PR #83, opened 2020-11-24
Mark HIPs 13 & 16 as deployed

georgica:


jamiew:
Thx, amending with some other links and information from #85

jamiew:
Ah sorry I mixed up which HIP this was about. Just rolled back my amended changes. HIP20 has been approved, but is not yet deployed; going to close as not applicable.

PR #91, opened 2020-12-10
HIPxx: Anchor Gateways

georgica:


jamiew:
Proposal looks solid to me. Will number and merge tomorrow morning 👍

jamiew:
I assume you prefer calling it Anchor Gateways instead of Golden Gateways?

georgica:
You can rename it to Golden Gateway since its more popular in the community

georgica:
i corrected a few misspelled words

cvolkernick:
also a fan of this, looks legit.

jamiew:
Say hello to HIP22. merging

lthiery:
@georgica I can work on specifying more clearly in the HIP, but I think that the "Golden Gateway" from a hardware specification boils down to a concentrator card that is RAK2287/RAK2247 compatible at its core. Do you agree with that?

The reason I think its important is that it may allow for eventual retrofits for operators who have any of the currently on-chain gateways.

lthiery:
oops i'll move convo to the issue now

PR #98, opened 2020-12-17
HIP19: more on key issuance and revocation

jamiew:
Diff: https://github.com/helium/HIP/pull/98/files

Rendered view: https://github.com/helium/HIP/blob/7129a3858dbe06ecb21a24b3436ab02c75416b34/0019-third-party-manufacturers.md

* Explicitly defers the details of a systen for actually issuing keys to manufacturer to a separate solution, through the HIP process or otherwise.
* Proposes the same HIP-style process for revoking manufacturer approval. Currently no way to remove hotspots from the network, and defers consideration
of that to a future proposal.

I believe that the two companies that have applied to be manufacturers should not be held in limbo just because we haven't finalized how they are going to get keys. I also don't think tracking keys on-chain is absolutely a firm requirement, since it could be quite involved.

### DIFF
--- a/0019-third-party-manufacturers.md
+++ b/0019-third-party-manufacturers.md
@@ -86,15 +86,19 @@ In order to join the blockchain, every hotspot requires an onboarding code. This
 
 Currently, RAK Wireless acquires codes from Helium Inc via a proprietary, manual process, with records of serial numbers or MAC addresses presumably kept by Helium Inc.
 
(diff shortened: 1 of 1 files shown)


cvolkernick:
agreed that on-chain issuance can be added later to "codify" what is currently an imperfect manual solution. As long as we are making improvements with each iteration.

Issue #87, opened 2020-11-25
HIP19: Approval Process For Third-Party Manufacturers


jamiew:
**Authors:** @jamiew (jamiedubs), @georgica, @philltran 
**Initial PR:** #86
**Start Date:** 2020-11-14
**Category:** Governance (Meta)

**Rendered view:**
https://github.com/helium/HIP/blob/master/0019-third-party-manufacturers.md

**Summary:**

This proposal seeks to lay out requirements for new third-party manufacturers to be approved by the community, and a process by which onboarding keys can be issued by Helium, Inc to those manufacturers.

shawaj:
Would be very interested in this for use with balenaCloud and Pi Supply / Nebra gateways 👍

PaulVMo:
I would really like to see more detail in the HIP on what minimum security requirements will be. I don't think this goes far enough and could lead to more, easily exploitable hotspots similar to RAK. Put another way, this does not seem to preclude RAKspots which are currently leading to significant growth in gaming. The HIP talks about favoring security but does not actually spell out any requirements as to what that means (e.g., hardware keys, non-removable storage, secure firmware). 

Down the line if and when there is better anti-gaming and/or different classes of gateways (e.g., "light gateways") this may be an acceptable set of requirements. However, those things do not current exist. At this point, I would prefer not to approve more manufacturers that are as insecure as the current generation of hotspots. To me, this does not provide enough clarity on those requirement and thus leaves the door open to more untrustworthy hotspots which are detrimental to the growth of the Helium network.

georgica:
The way i see it is:
1) lock down the firmware as much as you can
2) Better to encrypt the key folder on the SD card (you can't really copy encrypted folders)
3) Make sure the decryption key/passphrase is hard to obtain

philltran:
I agree with @georgica. We should only require due diligence in securing the hardware.

georgica:
to add to the above:
- Root user should not accessible
- Malicious devices cannot be plugged in, USB or otherwise
- If you do allow HID devices make sure user password cannot be reset easily.
(One example being using a flavor of linux booting into single user mode and resetting root password)

jamiew:
Baking some of the requirements mentioned here into the HIP in #90 (WIP)

shawaj:
@ryanteck and I were discussing this earlier today. 

[...]

PR #99, opened 2020-12-20
HIP19: mark as approved

jamiew:
Following onto #87 

Via Discord discussion, informal polls, the most recent community calls, and private discussions with folks not represented in those other forums, the Helium community have expressed overwhelming support for approving this proposal and moving forward w/ evaluating the two current applicants, Nebra Ltd & SyncroB.it

Addressing the remaining concerns: 

* Helium Inc have indicated their support and willingness to accept onboarding keys from approved manufacturers, and developing a process by which the issuance of those keys can be made public. 
* The community & manufacturers have both agreed that it's OK to approve this proposal this system is still being developed
* Lastly, it is agreed that adding the ability to revoke existing keys should be addressed in a separate proposal.

On behalf of the DeWi and the Helium community I'm calling this proposal ✅ approved

### DIFF
--- a/0019-third-party-manufacturers.md
+++ b/0019-third-party-manufacturers.md
@@ -5,6 +5,7 @@
 - Category: Meta
 - Original HIP PR: https://github.com/helium/HIP/pull/86
 - Tracking Issue: https://github.com/helium/HIP/issues/87
+- Status: Approved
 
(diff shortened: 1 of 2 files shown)


shawaj:
Great news 👍

PR #102, opened 2020-12-22
Update 0023-decouple-consensus-from-gateways.md

maco2035:
Change XXXX to 23 for the hip number assigned.


jamiew:
Thanks, merging

PR #104, opened 2021-01-01
Transfer Percentage of Hotspot

ericmheilman:
Submitting this PR early as I believe the current version is satisfactory.

Ed: rendered view of latest revision https://github.com/ericmheilman/HIP/blob/master/0024-transfer-percentage-of-hotspot.md


jamiew:
Looks great to me. Merging for discussion.

PR #108, opened 2021-01-09
Update 0007-managing-hip-process.md

georgica:
Updated communication channels from Slack and forum to Discord since it was referenced in HIP-19


jamiew:
Thank you!

PR #109, opened 2021-01-11
Update 0024-transfer-percentage-of-hotspot.md

ericmheilman:
Made a few clarifications in the summary.


mfalkvidd:
If the transaction has an HNT amount, where will that HNT amount go? I assume it will go from the original owner to the new owner, but it would be good to explicitly state the from and to parties.

jamiew:
PR by a named author. Auto-merging

*I'm not a bot, but I should be

PR #112, opened 2021-01-12
Update README.md

georgica:
Added HIP-25 Validators


evanmcc:
looks like this overlapped with a different PR, thank you.

PR #116, opened 2021-01-22
Update 0023-decouple-consensus-from-gateways.md

cvolkernick:
removed some template text leftover

### DIFF
--- a/0023-decouple-consensus-from-gateways.md
+++ b/0023-decouple-consensus-from-gateways.md
@@ -42,15 +42,6 @@ Community debate / discussion will be solicited via Discord in the respective hi
 # Detailed Explanation
 [detailed-explanation]: #detailed-explanation
 
-- Introduce and explain new concepts.
-
-- It should be reasonably clear how the proposal would be implemented.
-
-- Provide representative examples that show how this proposal would be commonly
-  used.
-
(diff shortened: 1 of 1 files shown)


cvolkernick:
@jamiew I'd also like to propose we possibly rename this and refocus the discussion around DPoS / staking extension of the more core technical changes, as HIP 25 has more or less covered the core changes originally described in the generic as "decoupling consensus from gateways".  I think we can consider that HIP as "approved" based on the preliminary vote from discord and move the conversations about overstake, slashing, DPoS / gateway involvement here.

Open to thoughts / suggestions here. :)

DGKSK8LIFE (approved):

DGKSK8LIFE:
> I'd also like to propose we possibly rename this and refocus the discussion around DPoS / staking extension of the more core technical changes, as HIP 25 has more or less covered the core changes originally described in the generic as "decoupling consensus from gateways". I think we can consider that HIP as "approved" based on the preliminary vote from discord and move the conversations about overstake, slashing, DPoS / gateway involvement here.

I believe we should wait until the community call on Wednesday before moving forward with refocusing the discussion.

cvolkernick:
That sounds reasonable @DGKSK8LIFE

jamiew (approved):

PR #117, opened 2021-01-24
HIP-xxxx: Payment Notes

cvolkernick:
This is a proposal I've watched get knocked around the discord in various forms for awhile now. Figured I'd make it an official proposal and gather feedback from the community.


jamiew:
Will review this evening and let you know if any changes needed prior to merging

cvolkernick:
Thanks @jamiew you the real MVP

PaulVMo:
Couple thoughts here:
1) If messages are on chain, I think you will not see a lot of adoption. For example, if I want to connect with a nearby hotspot owner, I would probably be looking to message them with my contact info to coordinate directly. However, I am unlikely to put my contact info on the public chain.
2) as an alternative, I could see this as a capability of the wallet app which can be handled off chain (I think there is a precedent for that with whatever messaging is used for the exchanging the partial transfer transactions). Almost all owners currently use the app. With it going open source, there is an opportunity to push for features to be added. The only drawback I see is that it arguably should be an opt-in feature which means you still may not be able to contact the person you want to.

jamiew:
Looks good enough to merge for discussion. Stand by for tracking issue

PR #127, opened 2021-01-30
On-chain HIP Voting

maco2035:
Proposed hip to change voting


philltran:
Here is a link to the [rendered view](https://github.com/helium/HIP/blob/541e4d8aaf16b05d57f6d4bb83ce47656e65a476/0027-HIP-Voting.md) https://github.com/helium/HIP/blob/541e4d8aaf16b05d57f6d4bb83ce47656e65a476/0027-HIP-Voting.md

Thanks for submitting this HIP. I think onchain voting will bring quantifiable results to future HIP discussions and further empower the people in the people’s network.

amirhaleem:
the HIP process today really encompasses several discrete steps, that are all relatively ineffeciently bundled together. but an attempt to break them out would be the following:

1 - discussion of the idea. this usually happens on discord or in private channels. but some attempt to rationalize the proposal, gain feedback on it, and discuss it in a larger group before proposing it in 'official' form. an example would be the "auditors" idea that I shared as a google doc to get feedback on. 
2 - drafting the HIP. this part is relatively easy once #1 has occurred, but is still time consuming. sometimes new ideas or suggestions surface here and the HIP is re-updated a few times before being submitted.
3 - technical feasibility. this is probably the most complicated part to assess, as there are not many community participants with deep enough technical knowledge to know the resources required, downstream effects or feasibility of implementing a given proposal.
4 - voting. this is today usually done through some combination of discord and community calls, with the final approval often coming in the community calls.
5 - implementation. to-date this has been handled entirely by the helium, inc team.
6 - testing. the implementation has to be tested in some form, either by simulator, optimally a testnet, or worst case a mainnet deployment with the option to rollback
7 - assessment. once the implementation has been deployed for a while have any negative effects been observed? does anything need changing? should it be altered or rolled back?

this HIP rightly suggests that step 4 is suboptimal. a straw-poll via discord or a community call is not the most efficient or equitable way to gain consensus on a given proposal. 

in other blockchain networks, using Bitcoin or Ethereum as an example that is similar in structure to helium, proposals are submitted in the form of code. a BIP in bitcoin is nearly always accompanied with the actual code required to make the change. the existing community and core development team then engage in a discussion about the merits or risks of the BIP, and if there is rough consensus that the code is good it is merged and included in the next release.

on-chain voting only solves point #4 in my list above. imagine for example a scenario where a proposal is suggested, 60% of hotspot owners agree, and the vote is now recorded on-chain. now what? what if it's not possible to implement the proposal the way it was described? what if implementing it would cause a vulnerability or security issue that was not considered in the proposal? who actually writes the code? 

in other words, what exactly is the value of recording the vote on-chain?

there are also simpler ways to do this if the desired goal is to provide a vote (the proposed methodology in this HIP is *extremely* complex). for example a proposal to split the number of DOT tokens in Polkadot simply created a few addresses representing the different voting options and had DOT holders send a nominal amount of tokens to whichever address they were voting for.

i have some other concerns with the proposal here, but they are less structurally important than this point.

i had written a HIP in the past that proposed a similar method, but widthdrew it as i think it introduces more complexity that the value it adds. at the end of the day an on-chain vote is only 1 of 7 (or more) steps outlined above, and on its own is relatively meaningless.

jamiew:
@maco2035 are you still interested in working on this? This HIP is in need of basic spelling and grammar edits before I think it could be merged for discussion. 

There is also the more substantive feedback mentioned by Amir and others, but I'm open to merging it as a kick-off for discussion about on-chain voting provided it was in a more readable state.

jamiew:
Closing as stale since a week without feedback. Please re-open if you're planning to continuing working on this

PR #130, opened 2021-02-09
amendment to hip19 for light hotspots

cokes518:

PR #132, opened 2021-02-17
Amend HIP19 for light gateway hotspots

cokes518:


cokes518:
amending the hip rather than introducing a new document as discussed here: https://github.com/helium/HIP/pull/130

lthiery:
I would suggest explicitly stating the security requirements (ECC608 or otherwise) do not change with light hotspots.

0019-third-party-manufacturers.md
```diff
@@ -79,7 +85,14 @@ Lastly, manufacturers are expected to provide:
...
 
 
+### Light Hotspot Requirements (Amendment)
+With the incoming launch of Validators (HIP25), the network no longer requires compute-intensive Hotspots to run consensus groups. This reduces the requirements for Hotspots such that they become "Light Hotspots". Light Hotspots are a new subclass of trusted Hotspots on the Network, able to participate in PoC, Witness, and create Challenges. In addition to PoC activities, they can transfer Data Packets. Light Hotspots will earn HNT for these activities. 
+
+Light Hotspots will **not** participate in Consensus and will be flagged as such on the Blockchain. They will not mine HNT for participating in Consensus.
+
+The reduced compute requirements will enable a new class of lower-cost Hotspots to enter the market.
 
+Light Hotspots will not be expected to pass the Audit and Compute requirements to run and participate in consensus.
```
abhay, line 95:
```suggestion
Light Hotspots will not be expected to pass the Audit and Compute requirements to run and participate in consensus. Their block absorption rate may be documented as a comparison to other blockchain following miners. 
```

abhay: Maybe worthwhile addition. We want to make sure Light Hotspots can keep up with the chain so they can perform PoC activities normally.

jamiew:
This looks fine to me, and would only have forward-facing impact.

Should we be applying the same process we use for new HIPs for amendments like this? Does this fall into "bug fix, doesn't need a full HIP" territory, or should we just always aim for slower approval of updates?

jamiew (approved):

jamiew:
Merging per my feedback last month. https://github.com/helium/HIP/pull/132#issuecomment-781401235

In future substantiave amendments should probably by their own HIPs, but this is not one of those

PR #135, opened 2021-03-10
fix LoRaWAN typo

zuyan9:
It should be LoRaWAN not LoRaWan


DGKSK8LIFE (commented):
Might want to add the "restricted trademark" symbol.

jamiew:
Thx, merging

PR #137, opened 2021-03-16
HIP-19 application for HeNet B.V. / LongAP

timcooijmans:
Rendered: https://github.com/HeNet/HIP/blob/feature/hip19-henet-application/0019-third-party-manufacturers/henet.md

### DIFF
--- a/0019-third-party-manufacturers/henet.md
+++ b/0019-third-party-manufacturers/henet.md
@@ -15,7 +15,7 @@ We have done projects where we had 100s of devices in the field managed by us an
 
 ## Product Information
 ### LongAP One
-**Status**: Fully functional production samples, CE testing passed.
+**Status**: Ready to start production, CE testing passed.
 
 **Expected release date**: Q2-Q3 2021
 
@@ -40,8 +40,35 @@ It's fully based on x64-hardware with reliable products such as Intel-interface-
 - Fully metal case
 - CE declaration of conformity available
 
-We are considering providing a variant with 4G/LTE connectivity instead of WiFi/LAN. However, this only becomes interesting once mining is moved from the gateways
-to validators.
+### LongAP Pro
+**Status**: Ongoing development
+
+**Expected release date**: Q3-Q4 2021 (after gateway-rs is ready)
+
+**Estimated price**: TBD. 
+
+LongAP Pro will be a light-gateway in an outdoor IP65 housing ready to be installed in the harshest conditions.
+
+**Specs**:
+- 800 MHz Dual-Core MIPS processor
+- 256 MB of RAM
+- Gigabit ethernet with PoE
+- LoRa concentrator (EU868 and US912 initially)
+- Fiberglass +3dBi LoRa Antenna with N-type connector
+- Fully metal case (pole and wall mount supported)
+- Powered over PoE
+
+Option LTE:
+- 4G/LTE Cat 1 interface
+- 2 Fiberglass LTE antennas with N-type connector
+- World-wide SIM with data installed that can be enabled in our dashboard
+
+Option WiFi:
+- WiFi interface
+- 2 WiFi antennas with N-type connector
+
+Option Solar:
+- Kit to run LongAP Pro on solar and batteries. 
 
 ## Customer Support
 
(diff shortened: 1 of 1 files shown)


jamiew:
Application received! Will review and share with wider community for
feedback

jamiew:
I can attest that @timcooijmans is who he says he is, and is a named owner of the entity [HeNet BV](https://drimble.nl/bedrijf/haps/48459658/henet-bv.html) via [Spurk Holding BV](https://drimble.nl/bedrijf/haps/48459658/henet-bv.html)

Carniverous19:
>Expected release date

April seems aggressive for delivering this product, April 30th is 6 weeks away from today.  This is time for completing bulk manufacturing of electronics and mechanicals, final assembly, shipping, Helium/DeWi audit, regulatory certification, and provisioning?  Can we expand on why this is likely to be successful?

> Company Information

Can we get more information on Velorum B.V. ? LongAP.com is just the logo, its mentioned that that is a spinoff which is fine but then we need to evaluate the parent company and their experience in volume manufacturing of gateway-like devices.  

*100s of devices in the field managed by us and operated over our own MVNO mobile network.*  Is a different scale than I think we are looking for from gateway manufacturers.  I'd expect gateway manufacturers to be able to produce and sell a few thousand hotspots a month to be a valuable addition to the community so will quickly get to 10,000+ gateways they will have to support.  Any  experience with higher volume?

 > Manufacturing Information

Need some better information on this.  High volume production on a gateway is more involved than sourcing available components.  There must be some sort of custom PCB fabrication (to get the secure element on something?).  There also must be some sort of enclosure to put these parts in? Is tooling made? Even if there was a warehouse full of parts these devices will need some sort of manual assembly.  If this is an off-the-shelf gateway kit that is just getting some minor retro-fit can you point to that device?

timcooijmans:
Please note that we didn’t just start developing, testing and producing today. We have been busy for quite some time and as indicated in our application, our solution is based on an existing board and existing radio-modules. We’ve used this board before for other applications. All the components are already tested and have CE approvals on their own (Yes we know CE doesn’t have complete modular approach like FCC but this makes the CE conformity assessment easier). We discussed with Helium representatives to open a HIP-19 application and to discuss the Helium/DeWi audit in parallel. While we have a complete working Gateway, we indeed are still pending the hardware-audit with Helium. So while we stated an expected (not guaranteed) release date, we adjusted it to be more broad, managing expectations, as this seems very sensitive at the moment in the Helium-community.

The only item specific for this solution in our bill-of-materials is indeed the PCB supporting the chip for securing keys. We have 1600 of those PCBs fully assembled scheduled for delivery next week. We can produce more in 2 weeks. And no our Gateway is not based on an existing LoRa Gateway. Our Gateway is targeted at the more professional users and has a different price-point then several larger Helium Gateway manufacturers. This means that we expect to sell in smaller volumes. Based on our experience with blockchains we believe that having a diverse offering of nodes of different kinds in a network is good for the ecosystem to reduce the reliance on vendors and make the network more resilient to implementation specific problems. For example we will be the first and only solution based on x86 hardware.

The upcoming migration of “validation” from the Gateways to dedicated Validators means that light-gateways will most likely become the Helium-community’s (and our) main focus around the end of this year, full gateways will become less interesting by then. Starting production in bulk-volumes (10k+ a month) of a full-gateway today is simply unrealistic for any new HIP-19 applicant in our opinion with light-gateways on the horizon. However we believe in Helium and want to help the Network grow in Europe today (or as soon as possible) and have a solution to do so in our hands. (Our current gateway will of course still work after the migration to light-gateways). Why not use it today and let the network grow?

Our involvement in large public and private blockchains and our connection to corporates within Europe means that we know what it takes to get big companies aboard. But sadly it also means that other than the public work one of us did previously at the Ethereum core-team and for Quorum, most of our work is under NDA and not public. However if required we can see if we can get private recommendations from our customers to DeWi and/or Helium for validation.

jamiew:
In recognition of:

- [x] Support from the Helium community, as evidenced by discussions and informal polling on GitHub and Discord
- [x] Identity verification by DeWi (above)
- [x] Regulatory compliance (CE approval)
- [x] Hardware/software/security review of prototype (confirmed by Helium Inc)

I am pleased to memorialize community consensus for approving **HeNet/LongAP** as a third-party manufacturer of Helium-compatible hotspots by marking this proposal approved ✅

PR #145, opened 2021-04-02
Dragino – HIP19 manufacturer application

dragino:
Rendered view: https://github.com/dragino/HIP/blob/master/0019-third-party-manufacturers/dragino.md


PharkMillups:
Thanks! Rendered view here:

https://github.com/helium/HIP/blob/cc49dc3c6e3eda39c6eecca54da4c5296897d925/0019-third-party-manufacturers/dragino.md

@jamiew and the DeWi will do a first pass and follow back up with initial questions.

jamiew:
We are transitioning HIP19 review process to a dedicated committee and a representative from that group will be in touch. I believe they have an initial meeting this week

PR #148, opened 2021-04-13
Update hip12 deployment status

vihu:
`assert_location_txn_v2` aka remote location assert is live!


jamiew:
LGTM, merging. Thanks for the PR

Issue #94, opened 2020-12-11
HIP22: DIY Concentrators


jamiew:
**Author(s):** @georgica, @lthiery, @Carniverous19 (para1)
**Initial PR:** #91
**Start Date:** 2020-11-16
**Category:** Technical

**Rendered view:**
https://github.com/helium/HIP/blob/master/0022-diy-concentrators.md 

**Summary:**

This proposal side-steps the inherent lack of trust in gateways by creating a special LoRa concentrator module that we can trust. Whereas today, RSSI, SNR and GPS timestamps cannot be inherently trusted, reports from these "Anchor Gateways" will be a root of trust for physical information on the network.

lthiery:
@georgica I can work on specifying more clearly in the HIP, but I think that the "Golden Gateway" from a hardware specification boils down to a concentrator card that is RAK2287/RAK2247 compatible at its core. Do you agree with that?

The reason I think its important is that it may allow for eventual retrofits for operators who have any of the currently on-chain gateways.

georgica:
@lthiery I agree with that it has to be integrated in the concentrator it's self. Like i stated in the HIP some modifications need to be done in order to ensure that it is un-temperable. 
Since RAK2287/RAK2247 are clones of Semtech's reference design with few additions, I have already started work on this board.

lthiery:
@georgica great, I thought we were aligned on that but I just wanted to make we were before trying to do a PR to the HIP :)

ctchadwick:
I am a new comer to Helium, but I really see the promise of the technology.  While it is good to expand the growth and deployment of the network, I do see the transient and speculative aspects of those acquiring hotspots simply for the mining rewards without any regard for the integrity of the network.  I'm a property owner with a longer horizon with regard to infrastructure, so I'd like to participate in the Anchor beacon part of the network.  Are there more details on this, perhaps some sort of registration/verification or list to get on for becoming an Anchor?  I apologize if this is not the place to address this.

lthiery:
@ctchadwick Welcome to the community! 

It's great to hear that the concept of this Helium Improvement Proposal (HIP) makes sense to you. However, as its name implies, the idea here is merely a proposal so far. At this phase, we are debating the proposal and trying to figure out if its policy we wish to adopt. Or if it may be improved from its current state to become policy we wish to adopt.

Therefore, there is no registration/verification; in fact, the methods proposed herein are attempts to provide trust without registration/verification but instead with a secure hardware and firmware provisioning process.

[...]

PR #160, opened 2021-04-22
HIP22: Switch focus to DIY Concentrators

lthiery:


timcooijmans:
While I think it's a good idea to allow DIY-concentrators I object to the claim that:
 >these are provably more honest witnesses
Running a normal "hotspot" could be as-secure as in fact it could use a very comparable architecture.

timcooijmans:
If you replace your "secure middleman" by a "secure processor" and replace the "PCIe goldfinger" by "internet" you have the current HIP19-approved setup. It's exactly the same security-architecture, but with a different components.

timcooijmans:
Sorry seems the discussion is in the issue. Wil post it there as well.

jamiew (approved):

jamiew:
Related to #94

PR #175, opened 2021-05-08
i2SOM - HIP19 third-party manufacturer application

Blackrose:
Please contact us with any questions at steve.chen@i2som.com

Rendered: https://github.com/i2som/HIP/blob/master/0019-third-party-manufacturers/i2som.md


ke6jjj:
As it stands, this proposal to manufacture _light_ gateways appears to be ok. I want to point out, however, that the proposal suggests that i2SOM plans to produce full on-chain miners by simply adding an ECC chip in later versions. The ECC chip will certainly be required to do so, but there's another hurdle that may be far more daunting: the CPU.

> Single core ARM Cortex-A7 **32bit** 800Mhz

There is currently no version of the miner which is known to work in 32-bit mode due to a yet-unfixed bug with one of the miner sub-components, `RocksDB`. This build issue will need to be triaged and fixed by someone before i2SOM can even attempt to create a 32-bit miner.  A 64-bit CPU is much more recommended.

Blackrose:
The current hardware can only run gateway-rs applications (https://github.com/helium/gateway-rs), not as a full miner. So, we have planned to use 64bit A35 CPU to implement the full miner hardware. Later, we will update this description to the file.

PR #180, opened 2021-05-22
HIP 31: Require manufacturers to include information about the legality antennas

timcooijmans:


abhay:
I agree with the spirit of this proposal (manufacturers providing guidance to customers in how to stay in compliance of local regulation), I don't think this rises to the level of an economic or technical improvement to the Helium network. 

My recommendation is that there are hardware manufacturer guidelines that the foundation set up for getting onboarding access (HIP19 can be updated to fit any new ideas) and also rules around maintaining access. Potentially the foundation may want to have rules around revocation (temporary or permanent) as well.

Curious on your thoughts @Scottsigel @Sophi?

timcooijmans:
I had some doubts whether this proposal was worth a HIP. The reasons why I opted for one anyway were:

- It introduces a first (consumer safe) way of "banning" manufacturers 
- It's very important for Helium to stay legal. The compensation using proof-of-coverage is a very smart invention that really is allowing the Helium network to grow. Sadly it also incentives (possibly unknowingly) breaking the law.

cvolkernick:
Wouldn't this be an amendment on HIP 19? That approach would seem to address both what @abhay and @timcooijmans have stated.

Sophi:
Agree wholeheartedly that Helium should stay legal.
I am not sure that it is practical to require manufacturers to provide
guidance -  who then verifies that what they provide is accurate and is up
to date? It seems that it would fit better with the onboarding process.
How often do manufacturers mislead customers by selling non-standard
antennas alongside hotspots?
I wonder if selling hotspots with non-standard antennas does in fact
increase coverage (I don't know and am looking into this).

timcooijmans:
I don't want to point fingers but several HIP-19 approved manufacturers are selling antennas that are almost impossible (meaning with normal length cables) to use legally in at least the EU without any warning.
And yes increasing coverage by using higher-gain antennas also increases earnings.

jamiew:
@timcooijmans per yesterday what do you think of idea to refactor this into an amendment to HIP19? I'd like to run through that new manufacturing committe for approval, but if they approved and community didn't have objections, It'd just become HIP19 v1.3 or so

timcooijmans:
@jamiew I wasn't aware the manufacturing committee was already formed. Can proposals be submitted to this committee via HIP? How does this actually work when the committee itself is working behind closed doors?


I also had some discussion with @shawaj if this should be enforced upon manufacturers or should be rewritten in some documentation provided by DeWi / Helium. 

My main concern is that most (especially non-professional) hotspot operators are not informed about the legality of using certain antennas (both sold by HIP19 manufacturers as bought elsewhere)

PapaOwl:
Is it impossible from a manufacturer side to enforce legality of the setup? I don't want to hear news of helium hotspot owners cuffed away, fined into oblivion.

I saw many people operate under the assumption, that nobody was gonna check. Which is true if we are talking 1-2 rule violations, but when we think about dozens or even hundreds of illegal operation in every city, then there will be pushback.

I am especially concerned, because at least for germany owning as little as 2 miners would already technically force you to open a business to run these profit machines and then you aren't in the realm of "private" people breaking the law a little bit, but a company increasing profit by breaching the law. This type of stuff can get out of hand if the opposing side wants to.

We could have legality checks by checking the connection strength of several nearby miners at the very least and eject miners based on that for a period.

jamiew:
> @jamiew I wasn't aware the manufacturing committee was already formed. Can proposals be submitted to this committee via HIP? How does this actually work when the committee itself is working behind closed doors? I mean how can I/we discuss this proposal with the committee, gather their feedback, and so on?

I have to defer to @Scottsigel and @Sophi for more on this, as I'm not actively involved with the committee. But I know it's still early so you should expect to hear more

Sophi:
HI @timcooijmans! 
We **just** got started with the committee, which is why there's little information.
Our second meeting will be on June 23rd and @Scottsigel or I will update you afterwards on how we'll all interact. 
For now, use email ;)
We are working on setting up a monthly manufacturing call with approved manufacturers and the committee - TBD, but by the end of the month (hopefully) all of these things should be in place.

kellybyrd:
Not sure if this is helpful, but I'll throw it out there. This feels like a similar problem vendors in R/C FPV (remote control first person view) hobby have, so maybe we can take a similar approach?  

Some background: In FPV hobby, a variety of frequencies are used for video and control (433Mhz, 900Mhz, 1.3Ghz, 2.4Ghz, 5.8Ghz are the most common). Just like with LoRA gear, there are transmit power limits and EIRP limits for the equipment being sold. But, most common FPV video transmitters are above the limit allowed by many regulatory domains unless you have an Amateur Radio license (HAM license in the US). To account for this most web shops have a warning on the bottom of all product pages that fall into this category. Examples:
 [GetFPV](https://www.getfpv.com/lumenier-tx5gpro-mini-200mw-5-8ghz-fpv-transmitter-with-power-switch.html)
> [...snip...]
> It is your responsibility to ensure that the use of this product meets the requirements imposed by your government's rules and regulations for RF devices. Do not purchase this product if you are unsure of the government requirements or are not able to comply with them. GetFPV cannot be held responsible for your actions if you purchase and/or use this product in violation of your government's regulations. By agreeing to our terms and conditions during the check-out process, you also agree to have read and to understand this notice.

[HobbyKing](https://hobbyking.com/en_us/skyzone-ts58500-32ch-500mw-mini-video-transmitter.html)
> Attention Customers
** The use and operation of this product in some countries may require a license based on local government restrictions.  Ensure you check your local government rules before operating this device. **

They are solving for a slightly different problem, but my point is that the customer warning is not very specific and just directs customer to ensure they meet the requirements of their regulatory domain. HobbyKing in particular was fined by the FCC a few years back because even though that had that warning, several video transmitters they sold used frequencies and power levels beyond what even an Amateur Radio license would allow.

mfalkvidd:
I share @timcooijmans fear that the People's network will get penalized when too many users (knowingly or unknowingly) use equipment that submits stronger signals than allowed by law.

Multiple manufacturers promote their highly directive antennas for use with their Helium hotspots, without any mention at all that the combination will likely not be legal. If required, I can show listings by multiple vendors but I would prefer to not single out any specific manufacturers.

Getting into a legal fight, or having equipment seized at customs would hamper the expansion of the network.

wolfenhawke:
This should be covered by FCC cert, or equiv per geography. The cert includes the antenna. So, really from a manufacturers perspective two things are required. 1. note FCC and other region certification, including cert number during shipments, 2. provide information of highest gain and pattern tested and certified which is an allowance by FCC part 15.204(c), requiring testing only with the highest gain antenna of each type that would be used with the transmitter at the maximum output power of that transmitter.  Any antenna of a similar type that did not exceed the antenna gain of tested antennas could be used without retesting.  Manufacturers would be expected to supply a list of acceptable antenna types with applications for equipment authorization. have warning that users are not advised to change antennas, unless they are professional installers conforming with regulatory requirements. 3. provide the warning noted above.

mfalkvidd:
Only following FCC is a bit narrow-minded, since Helium exists outside the US. But following FCC would be a good start.

> have warning that users are not advised to change antennas, unless they are professional installers conforming with regulatory requirements. 3. provide the warning noted above.

Currently, the above warning is generally not provided. Instead, vendors often explicitly recommend buying an antenna that exceeds the allowed EIRP.

PR #196, opened 2021-06-04
clarify implementation of net emissions

evanmcc:
@tjain-mcc @jmfayal 

Responding to questions on discord, I'm proposing this clarification to the implementation of net emissions with respect to halving events.


rawrmaan:
I think this is a good clarification.

jamiew (approved):

jamiew:
🔨

Wazneh:
When does the net emissions kicks in ?
Is it working now ? And supplying the burned hnt ?
Or will it work after the dc burned is bigger than the minted ?

PR #199, opened 2021-06-08
Update 0017-hex-density-based-transmit-reward-scaling.md

dsalis72:
DexDensity_Unclipped seems to be a type as the variable in the formula above is HexDensity_Unclipped


jamiew:
Blessed are the typo fixers

Thx 

cc @Carniverous19

PR #216, opened 2021-06-24
Clarify that DeWi KYC/B happens at beginning of approval process

Beniben:


jamiew:
Afaik this is actually true now, so I'm going to fix the `2-4` to say `2-5` and go ahead and merge

PR #226, opened 2021-07-01
Updates to HIP32

PapaOwl:


jamiew (approved):
Helped sort out some merge commits, lgtm now

PR #256, opened 2021-08-06
HIP19 - Amendment - Refunds

mrfizzy99:
Manufactures/Vendors MUST offer refunds up until the day of shipping if payment is made via traditional Creditcard or Paypal.  

Plain and simple

### DIFF
--- a/0019-third-party-manufacturers.md
+++ b/0019-third-party-manufacturers.md
@@ -134,7 +134,7 @@ Both Helium and RAK hotspots are (presumably) running the same miner software as
 ## Customer Refunds
 
 All manufacturers must offer an avenue of a refund in the orginal currency paid (excluding crypto) up until the item has been shipped to customer.  
-This encompasses the traditonal creditcard or paypal options for customers.  Wire Transfers for B2B may remain as-is per the sellers policies.  
+This encompasses the traditional creditcard or paypal options for customers.  Wire Transfers for B2B may remain as-is per the sellers policies.  
 
 
 ## How To Apply
--- a/0019-third-party-manufacturers.md
+++ b/0019-third-party-manufacturers.md
@@ -130,6 +130,13 @@ The official Helium mobile app is the primary interface for onboarding a new hot
 
 Both Helium and RAK hotspots are (presumably) running the same miner software as available on GitHub, but software updates are handled by a proprietary process. Additionally, there is remote access that can be used by Helium staff to diagnose and resolve issues. Both of these are closed-source and undocumented. If not open-sourced, third-party manufacturers would need to develop their own provably-reliable alternative.
 
+
+## Customer Refunds
+
+All manufacturers must offer an avenue of a refund in the orginal currency paid (excluding crypto) up until the item has been shipped to customer.  
+This encompasses the traditonal creditcard or paypal options for customers.  Wire Transfers for B2B may remain as-is per the sellers policies.  
+
+
 ## How To Apply
 
 Interested manufacturers should:


capjbadger007:
Seems like a good basic service any maker should offer. Also a bit of carrot/stick for those that abuse the customers' trust.

oosui:
🤣 Shouldn't Manufactures/Vendors obey the law?

capjbadger007:
> 🤣 Shouldn't Manufactures/Vendors obey the law?

What law are you talking about?

oosui:
> > 🤣 Shouldn't Manufactures/Vendors obey the law?
> 
> What law are you talking about?

There should be an agreement between the seller and you when ordering the product. They should refund you according to the agreement.

capjbadger007:
> > > 🤣 Shouldn't Manufactures/Vendors obey the law?
> > 
> > 
> > What law are you talking about?
> 
> There should be an agreement between the seller and you when ordering the product. They should refund you according to the agreement.

I agree, but there isn't right now. In fact, the order page for the SEEED Sensecap M1 says no refunds, no order cancellations.

HeyyPhil:
100% agree with this but feel there is more amendments to be added.

capjbadger007:
> 100% agree with this but feel there is more amendments to be added.

What did you have in mind?

oosui:
> > > > 🤣 Shouldn't Manufactures/Vendors obey the law?
> > > 
> > > 
> > > What law are you talking about?
> > 
> > 
> > There should be an agreement between the seller and you when ordering the product. They should refund you according to the agreement.
> 
> I agree, but there isn't right now. In fact, the order page for the SEEED Sensecap M1 says no refunds, no order cancellations.

hmm, if so, just don't buy it. I don't think there is a good way to solve this issue. It's same everywhere, even on ebay, amazon. It's the choice for sellers to protect themselves. What will you do if you are the seller? Accept refund without agreement? Because it's preorder, the risk is too high. Suppose you put a one billion dollar order and they started to make hotspots for you. Then you cancelled the order before shipping, who will pay for the cost of these hotspots?

capjbadger007:
> > > > > 🤣 Shouldn't Manufactures/Vendors obey the law?
> > > > 
> > > > 
> > > > What law are you talking about?
> > > 
> > > 
> > > There should be an agreement between the seller and you when ordering the product. They should refund you according to the agreement.
> > 
> > 
> > I agree, but there isn't right now. In fact, the order page for the SEEED Sensecap M1 says no refunds, no order cancellations.
> 
> hmm, if so, just don't buy it. I don't think there is a good way to solve this issue. It's same everywhere, even on ebay, amazon. It's the choice for sellers to protect themselves. What will you do if you are the seller? Accept refund without agreement? Because it's preorder, the risk is too high. Suppose you put a one billion dollar order and they started to make hotspots for you. Then you cancelled the order before shipping, who will pay for the cost of these hotspots?

We set the rules. If they want into this market, they play by the rules we set.

mrfizzy99:
> > > > > 🤣 Shouldn't Manufactures/Vendors obey the law?
> > > > 
> > > > 
> > > > What law are you talking about?
> > > 
> > > 
> > > There should be an agreement between the seller and you when ordering the product. They should refund you according to the agreement.
> > 
> > 
> > I agree, but there isn't right now. In fact, the order page for the SEEED Sensecap M1 says no refunds, no order cancellations.
> 
> hmm, if so, just don't buy it. I don't think there is a good way to solve this issue. It's same everywhere, even on ebay, amazon. It's the choice for sellers to protect themselves. What will you do if you are the seller? Accept refund without agreement? Because it's preorder, the risk is too high. Suppose you put a one billion dollar order and they started to make hotspots for you. Then you cancelled the order before shipping, who will pay for the cost of these hotspots?



Well, you don't do "Pre-Orders" for a product that won't be created until 3 months from that order.  Simple as that. :)

oosui:
> > > > > > 🤣 Shouldn't Manufactures/Vendors obey the law?
> > > > > 
> > > > > 
> > > > > What law are you talking about?
> > > > 
> > > > 
> > > > There should be an agreement between the seller and you when ordering the product. They should refund you according to the agreement.
> > > 
> > > 
> > > I agree, but there isn't right now. In fact, the order page for the SEEED Sensecap M1 says no refunds, no order cancellations.
> > 
> > 
> > hmm, if so, just don't buy it. I don't think there is a good way to solve this issue. It's same everywhere, even on ebay, amazon. It's the choice for sellers to protect themselves. What will you do if you are the seller? Accept refund without agreement? Because it's preorder, the risk is too high. Suppose you put a one billion dollar order and they started to make hotspots for you. Then you cancelled the order before shipping, who will pay for the cost of these hotspots?
> 
> Well, you don't do "Pre-Orders" for a product that won't be created until 3 months from that order. Simple as that. :)
> 
> Also a "One-Billion" dollar order would a B2B and would be a Wire Transfer. Imagine giving credit card / Paypal 3% of that transaction lol.



> > > > > > 🤣 Shouldn't Manufactures/Vendors obey the law?
> > > > > 
> > > > > 
> > > > > What law are you talking about?
> > > > 
> > > > 
> > > > There should be an agreement between the seller and you when ordering the product. They should refund you according to the agreement.
> > > 
> > > 
> > > I agree, but there isn't right now. In fact, the order page for the SEEED Sensecap M1 says no refunds, no order cancellations.
> > 
> > 
> > hmm, if so, just don't buy it. I don't think there is a good way to solve this issue. It's same everywhere, even on ebay, amazon. It's the choice for sellers to protect themselves. What will you do if you are the seller? Accept refund without agreement? Because it's preorder, the risk is too high. Suppose you put a one billion dollar order and they started to make hotspots for you. Then you cancelled the order before shipping, who will pay for the cost of these hotspots?
> 
> Well, you don't do "Pre-Orders" for a product that won't be created until 3 months from that order. Simple as that. :)
> 
> Also a "One-Billion" dollar order would a B2B and would be a Wire Transfer. Imagine giving credit card / Paypal 3% of that transaction lol.

Yes, you can set a rule that makers must have inventories before they start to sell them. 1 billion is just an example here ^_^, actually it can be 1 million orders from multiple buyers.

ganey:
In the UK, the Consumer Rights Act 2015 might have some coverage here on refunds (even if paid in crypto).

There's a section here on digital goods requiring refunds: https://www.legislation.gov.uk/ukpga/2015/15/section/44/notes

> ...
However, digital currencies (or cryptocurrencies) that can be used in a variety of transactions with a number of traders, and exchanged for real money, are much more akin to real money (e.g. bitcoins). Where the consumer uses these types of digital currency to pay for digital content, the trader can (and must, unless the consumer agrees) repay the consumer in the digital currency. The trader cannot charge the consumer a fee for the payment of the refund.

There's also this bit in section 20, failing refund you can pursue damages.

Section 20 of the Consumer Rights Act, 114: https://www.legislation.gov.uk/ukpga/2015/15/notes/division/3/1/3/4/2

> 114. If money was not used to pay for the goods, under subsections (11) and (12), the consumer is not entitled to money back but the “refund” would be a return of whatever the consumer gave in exchange for the goods (see examples below). If this cannot be returned to the consumer due to the nature of the exchange, then, under subsection (18) and (19), the consumer may not demand a refund but may pursue a damages claim.

I'm not clear on what happens if the value of a coin goes up or down since the original transaction?
If the price of a coin increases from the time you made the purchase and you received less as a refund, I guess it's possible you could put in a claim for those losses, after all the receiving company now has to pay Capital Gains Tax on an increases in value when they cash out.

Also, there's a _possibility_ to claim for "loss of enjoyment" due to late deliveries, but I'm not sure how likely it is to actually win that case.

On a separate note, I think I read in discord someone has raised a case against a vendor with the Advertising Standards Agency (ASA) here in the UK due to a vendor being massively late on delivery estimates.

Zaperking:
> Manufactures/Vendors MUST offer refunds up until the day of shipping if payment is made via traditional Creditcard or Paypal.
> 
> Plain and simple

Shouldn't matter what the form of payment as. Whether it is stablecoin or HNT, too. Approved manufacturers should pledge to offer refunds for reasons such as batch being delayed, etc etc.

wolfenhawke:
In this environment refunds are a non-starter. Leadframes are out by 12+ months, wafers are out by 6+ months. What this means is that Semtech puts in an order to TSMC to get some chips to make their radio; and to their packaging house to make modules and the lead time could be 1 year +. I am living this reality as I work in the semiconductor industry. Now go up the silicon stack. Rak puts in an order for you. They get priority (what little there is) based on volume of orders and so will stack requests to put in a minimum order for silicon parts to ensure they will get (still super crappy) delivery times. Almost ALL semiconductor orders now are NCNR (non cancellable, non returnable). We (my employer) have that rule from TSMC, and we (as a semiconductor manufacturer) pass that along to our customers. 
Now Joe Smith wants to cancel 1 month into the order. RAK will be left holding the can. 
The result: you will see prices become very dynamic and just keep going up. That freed up HS that Joe Smith cancelled? Resell it for 10%+. This will keep happening, and our costs will go up very dynamically.
Some points as noted above. 
1.The manufacturers are not "sitting" on your money for months. They are already paid for the product, they are also waiting. So that may follow the UK law, "If money was not used to pay for the goods,", well it was.
2.Best in this environment is for a clear agreement from the seller and the buyer. Then buyers can decide up front if they are ok with that.

mrfizzy99:
> In this environment refunds are a non-starter. Leadframes are out by 12+ months, wafers are out by 6+ months. What this means is that Semtech puts in an order to TSMC to get some chips to make their radio; and to their packaging house to make modules and the lead time could be 1 year +. I am living this reality as I work in the semiconductor industry. Now go up the silicon stack. Rak puts in an order for you. They get priority (what little there is) based on volume of orders and so will stack requests to put in a minimum order for silicon parts to ensure they will get (still super crappy) delivery times. Almost ALL semiconductor orders now are NCNR (non cancellable, non returnable). We (my employer) have that rule from TSMC, and we (as a semiconductor manufacturer) pass that along to our customers.
> Now Joe Smith wants to cancel 1 month into the order. RAK will be left holding the can.
> The result: you will see prices become very dynamic and just keep going up. That freed up HS that Joe Smith cancelled? Resell it for 10%+. This will keep happening, and our costs will go up very dynamically.
> Some points as noted above.
> 1.The manufacturers are not "sitting" on your money for months. They are already paid for the product, they are also waiting. So that may follow the UK law, "If money was not used to pay for the goods,", well it was.
> 2.Best in this environment is for a clear agreement from the seller and the buyer. Then buyers can decide up front if they are ok with that.

Joe Smith is a customer ordering 1, is that going to ruin RAK?   
I don't understand you "RAK will be left holding the can", what can?  There will be other buyers, RAK (and said other companies) believe in this product for the long hall, right?  They wont just stop producing them if orders stop flowing in, right?    They won't need extra units in the future for when legitimate RMAs flow in, right?  

If I go to any retail store, do I just see rows of empty shelves with signs that say "Pre-Order; NCNR"?   
Damn, any retail store with inventory must be losing money just by having stock of something that hasn't been bought yet.

wolfenhawke:
@mrfizzy99 No, 1 customer return is not going to ruin RAK. While you may know there is a charge for using a credit card, there is another charge for the vendor to rebate it. There is someone manually "restocking" even if it's digital. But, many people throwing cigarettes, many people changing their mind will ruin a good thing. Again, nothing that cannot be solved by the subsequent sale going up in price to account for the extra cost. So, RAK or any other manufacturer will not be ruined, particularly since there is such high demand, but their costs will go up and their money is tied up with the vendor. So greater demand will drive even higher prices faster to account for the lost revenue of changing minds, and borrowing costs to cover returns.
And yes, stores measure profit by "turns". Not just selling something, but how often their stock sells per month. So, something that is sitting on the shelf longer, is "losing money". It's the shelf space that has value, not the thing on it to a store owner.

ganey:
> And if a card issuers starts to see a lot of chargebacks appear for a single merchant.. well i'm sure there is more discussions on what they are doing to get these chargebacks and make a case on why they (the company) is not fraudulent.

I imagine this is something Nebra is struggling with, the massive silence from them and the misleading estimates given by them are causing more and more people to just put in chargebacks or get refunds via PayPal.

There's some people with dead Nebra's not getting support for over two weeks, they're more than entitled to just put in chargebacks.

shawaj:
We aren't struggling with chargebacks precisely for the reasons @mrfizzy99 states - we offer full refunds up until the time of shipping. 

Seems like a good policy IMO

ganey:
@shawaj I noticed that Nebra has previously stated it doesn't offer crypto refunds.

I was wondering if you can clarify which part of the 'distance selling regulations' or the 'consumer rights act 2015' exempts Nebra from providing the refund? 

I do accept that potentially the refund would not be for the original HNT amount due to the volatile nature of crypto (similar to if a user paid in alternative fiat currency at point of sale), I haven't had any clarification on this yet as to if it must be the same amount of HNT as the original transaction or not.

Here's a few links that covers the regulations: 

[UK Gov on distance selling refunds](https://www.gov.uk/accepting-returns-and-giving-refunds)

[Consumer Rights Act - 114 alternative payment methods](https://www.legislation.gov.uk/ukpga/2015/15/notes/division/3/1/3/4/2)

[Consumer Rights Act - 215 digital refunds in cryptocurrencies](https://www.legislation.gov.uk/ukpga/2015/15/section/44/notes)

ganey:
> > 🤣 Shouldn't Manufactures/Vendors obey the law?
> 
> What law are you talking about?

I didn't notice this bit before, but UK distance Selling regulations gives consumers who buy goods online up to 14 days after delivery for a no questions asked full refund (minus delivery costs). 

Doesn't look like it matters what you paid with either (based on consumer rights act).

DSR also says this:
`It’s illegal to display any notice that deliberately misleads consumers or deceives them about their rights, for example a sign that says you do not accept returns or offer refunds`

cvolkernick:
> @shawaj I noticed that Nebra has previously stated it doesn't offer crypto refunds.

@ganey I'm wondering now that this is mentioned, what is the reasoning behind no refunds of crypto purchases [in general]?

If following are all true...

1. Invoiced / billed total in local currency (e.g. USD/GBP) is documented & agreed upon by both parties.
2. Exchange rate & total crypto payment required to settle the invoice balance due is documented & agreed upon by both parties.
3. Sending / receiving addresses & transactions between them are documented publicly on-chain.

...then why is it that the fiat equivalent cannot be be refunded? Or frankly even just the original amount of the respective cryptocurrency just be re-issued to the account that sent it. That option removes any question of exchange rates from the equation.

Pardon my ignorance here if I'm missing something. I had just found that somewhat odd. TIA!

wolfenhawke:
I can see "the fiat equivalent cannot be be refunded" being more easily supported. Problem with crypto refund is most companies are in the game to make a profit now. Crypto with it's high volatility would require a separate holding of it as a corporate fund to allow disbursement. This fund would then be untouchable funds. It's another cost center then (lost opportunity cost). So, thinking you are doing a favor by accepting crypto, as a company, starts to get more complicated with things like refunds.
Also, as people start to pull their refunds from 3rd parties, say PayPal, the manufacturer loses in the net. There is a charge for the original credit transaction, and they have another charge put on them for the refund. PayPal gets profit on both ends.

ganey:
@cvolkernick Thats a good point, i'd only thought of it in the case of ordering miners, but yeah **any** UK company accepting crypto will have to offer refunds of some form.

Legally whenever a company transfers out HNT (burning or cashing out etc) they have to declare any increase/decrease in value as part of Capital Gains Tax. I'm not sure how it works for really old accounts with companies like MtGox involved with no available transaction history

Accepting crypto in the UK seems to me like more of a headache than fiat because it's so volatile and you have to keep records of everything. Even if a company is dissolved, the HNT is an asset, and owner of the wallet holding the HNT would be legally bound to pay Capital Gains Tax when the currency is moved, that could be rather significant in say 10 years time.

jamiew:
@mrfizzy99 would you be willing to present this idea for discussion at next DeWi community call? this upcoming Weds @ noon ET. https://dewi.org/community-call – I was not sure about your Discord username or would ping you there

PapaOwl:
> 
> 
> In this environment refunds are a non-starter. Leadframes are out by 12+ months, wafers are out by 6+ months. What this means is that Semtech puts in an order to TSMC to get some chips to make their radio; and to their packaging house to make modules and the lead time could be 1 year +. I am living this reality as I work in the semiconductor industry. Now go up the silicon stack. Rak puts in an order for you. They get priority (what little there is) based on volume of orders and so will stack requests to put in a minimum order for silicon parts to ensure they will get (still super crappy) delivery times. Almost ALL semiconductor orders now are NCNR (non cancellable, non returnable). We (my employer) have that rule from TSMC, and we (as a semiconductor manufacturer) pass that along to our customers.
> Now Joe Smith wants to cancel 1 month into the order. RAK will be left holding the can.
> The result: you will see prices become very dynamic and just keep going up. That freed up HS that Joe Smith cancelled? Resell it for 10%+. This will keep happening, and our costs will go up very dynamically.
> Some points as noted above.
> 1.The manufacturers are not "sitting" on your money for months. They are already paid for the product, they are also waiting. So that may follow the UK law, "If money was not used to pay for the goods,", well it was.
> 2.Best in this environment is for a clear agreement from the seller and the buyer. Then buyers can decide up front if they are ok with that.

You know that is the risk of companies and not customers and the law is quite clear on that. The market regulates these type of things itself . This isn't a queston if they should allow this, it is mandated for consumers to be protected this way in the EU and it doesn't matter if your chips are hard to get or your fabric is super rare or if there is a lithium shortage that drives the prices up.

That is the risk you take as business, when you start a project and you are free to opt out and produce batches you are more comfortable with or take another project altogether. Seeed managed their production in under 12 months and even under 6 months so clearly the market has ways.

TSMC is also not the only one that makes them non refundable. Tons of B2B orders are standing and cannot simply be voided. That is why we separate consumer law and b2b law.

shawaj:
@ganey to be honest, I'm not a lawyer so I don't really want to comment on legal issues over and above saying that we have taken advice from our lawyers and they have said what we are doing is ok. Whether they are correct or not is subjective I guess and probably isn't clear cut... As with a lot of things legal.

However, I think relying on the law of any country is pointless here. In part because things like DSR and CCR are only (easily?) enforceable in country - i.e. UK to UK sale for example. Selling internationally it becomes much more complicated. But also - the laws don't really matter in the context here... as this hip amendment is essentially seeking to go above what is defined in law. And for jurisdictions where the law is already more strict than what is defined here then the law already prevails anyway.

The other thing worth noting is that the vast majority of people do not actually want a refund - they want their hotspots quicker. All manufacturers currently are having supply issues and long lead times - some (read Nebra) have also had production issues on top however within the next 6 months to a year supply should open up considerably - for example I know for a fact that Semtech and Microchip (ATECC chips) are putting considerable extra resources into huge increases in production (we talk to them weekly at least), a lot of which is due to the huge demand in helium hotspots. Additionally there is efforts to add additional chip suppliers with things like [TPM support](https://github.com/helium/miner/pull/940) etc. This should go a pretty long way to opening things up and getting to a point where there is less of a gap between supply and demand at which point market forces will take over more readily anyway and consumers will have more choice etc. There has been a huge collection of issues that has led to this huge squeeze but things are definitely starting to ease (if you have scheduled chip orders already in place certainly).
 
I guess what all of this is to say is that the barriers to entry for new manufacturers should not be placed too high.... Because competition and more manufacturers is a good thing for everyone. It's a careful balance though for sure as protecting users with ability to refund before shipping is also very important and there doesn't seem to be any down side to it.

With regards to crypto refunds, again maybe the stipulation could be that vendors provide "at least one payment method which is refundable"... In other words if you only accept crypto then crypto is refundable. If you accept cards too you can have cards refundable but not crypto as long as it's clearly stated that is the case.

@PapaOwl Seeed is an outlier because they are already a huge reseller of Raspberry Pi so had huge stock already available and on order. And they were already launching the exact product for non Helium use cases. So it's not easily comparable to others who started from nothing. However, it certainly is different in the sense that they are selling from stock so refunds are less important in that scenario. Maybe a stipulation could be made that "pre-orders have to be refundable" whereas if they are from stock it could be more lenient. But seeing as there is huge demand, even if someone like Seeed selling from stock offered refunds I doubt many people would want refunds and it would only be an option for unopened and unused goods anyway (i.e. not onboarded)

666alawlessman777 (approved):

jamiew:
Closing stale issue. If this is still being worked on, please let us know.

Additionally, regulation of hotspot manufacturers is now handled by the dedicated MOC: https://github.com/dewi-alliance/hotspot-manufacturers

PR #259, opened 2021-08-12
Move HIP19 manufacturer applications to their new home

jamiew:
Specifically, here: https://github.com/dewi-alliance/hotspot-manufacturers

Still WIP but warming this up

PR #265, opened 2021-08-18
change security part

spalish:
We will protect and encrypt storage of the swarm_key by ECC608 chip.
We are willing to submit a prototype for audit, and sharing those audit results publicly.


jamiew:
Hi, per #259 all manufacturer applications now live in https://github.com/dewi-alliance/hotspot-manufacturers -- please submit updates there. cc @Sophi

PR #269, opened 2021-08-28
HIP draft: omni-protocol PoC on Helium Network

zer0tweets:
Publishing HIP 37: Omni-Protocol PoC on Helium Network per earlier discussions in the community meeting and various discord channels.


mfalkvidd:
For the unlimited plan gaming
> we propose to address the above by implementing a chain variable that would allow setting an upper limit on data transfer for each UE <> CBRS on a per operator basis.

Should we specify what happens when an UE reaches the limit? Are they denied access entirely? Or capped to a low speed?

Is the UE usage reset at any point (1st day of a new month for example)?

Is this a single chain variable, or does each operator have their own value?

jamiew:
Proposal looks great. Making some minor formatting fixes and then officially numbering and merging for discussion

jamiew:
Please direct future questions and comments to this HIP's new tracking issue, https://github.com/helium/HIP/issues/271

Authors – please tag #271 in future PRs

PR #272, opened 2021-09-06
Create 0038-validator-oracles.md

cvolkernick:
HIP proposing that staked validators be allowed to opt-in as price Oracles & submit price data.

### DIFF
--- a/0038-validator-oracles.md
+++ b/0038-validator-oracles.md
@@ -1,0 +1,110 @@
+# HIP Template
+
+- Author(s): @cvolkernick
+- Start Date: 2021-09-06
+- Category: Economic / Technical
+- Original HIP PR: <!-- leave this empty; maintainer will fill in ID of this pull request -->
+- Tracking Issue: <!-- leave this empty; maintainer will create a discussion issue -->
+
+# Summary
+[summary]: #summary
+
+From Helium Docs:
+
+>HNT converts to Data Credits at a rate pegged to $.00001 , and so the blockchain requires a canonical HNT/$USD price for this conversion. Beginning in June 2020, the Helium blockchain started using a system of decentralized price oracles to supply the $USD to HNT price used for on-chain for burn transactions. (This system is inspired by the Maker Foundation's Oracle usage.)
+>
+>To find this price, nine oracles (***this number may change in the future***) periodically submit HNT/$USD prices. Once the blockchain has enough new price data, it will calculate a new HNT/$USD price that will remain valid until enough new, valid oracle inputs are submitted, triggering a new price revision.
+
+This HIP proposes not only changing this number, but also expanding the pool of those who are able to make submissions. It proposes empowering any staked validators with the existing `price_oracle_submission` & allowing them to commit price oracle submissions as chain-blessed Oracles.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


philltran:
Here is a [link](https://github.com/cvolkernick/HIP/blob/94c86f987b2b898de1dccd134d271376aab18e24/0038-validator-oracles.md) to the rich text view - https://github.com/cvolkernick/HIP/blob/94c86f987b2b898de1dccd134d271376aab18e24/0038-validator-oracles.md

abhay:
I don't hate the idea because it does dramatically increase the size of the potential pool of oracle prices. Unfortunately, there's no limitations here on who the oracles are. In theory, the account with 300+ validators could take over price prediction very easily. 

The current system, even though has flaws, at least protects against Sybil attacks.

cvolkernick:
>In theory, the account with 300+ validators could take over price prediction very easily.

Are there measures that can be taken to prevent this...? E.g. limit to a certain # of submissions per period (blocks preferably), or attach a marginal cost of some kind? It seems inherently cost-prohibitive to stake 10K HNT 300x to manipulate the Oracle price if you can't even access your stake for 5 months.

Or perhaps there needs to be some minimum threshold in terms of submission diversity or number of submissions for the price to update (maybe in addition with the submission rate limiting or cost mentioned above)?

I don't necessarily question your point in and of itself, rather attempting to address the concerns you point out with creative solutions.

abhay:
Open to being creative.

Adding a limit is possible but it doesn't solve the issue of a plurality of submissions coming from a single _real_ entity. Splitting by account also causes similar issues so submission diversity isn't something that can be done on chain.

cvolkernick:
Understood; I suppose my outstanding question is, is it even plausible to do that to begin with if you are inherently required to stake 10K HNT (currently ~$215K) for _one_ validator? You would have to spend $1M just to submit 4 data points. With a current pool size of 2K+ that seems impractical IMO.

As a related note, is there any particular reason why an oracle-based pricing was chosen as opposed to pulling in market price data from exchanges, etc? Pulling data from the market almost seems to be a more practical way of obtaining this information that attempting to derive it from a manual process.

abhay:
> Understood; I suppose my outstanding question is, is it even plausible to do that to begin with if you are inherently required to stake 10K HNT (currently ~$215K) for _one_ validator? You would have to spend $1M just to submit 4 data points. With a current pool size of 2K+ that seems impractical IMO (assuming that a significant enough % of staked validators are participating, of course).

The monetary burden is an important one but why is the wallet with 300+ validators more important than the one with 1? Comes back to my earlier point about Sybil resistance. 

> As a related note, is there any particular reason why an oracle-based pricing was chosen as opposed to pulling in market price data from exchanges, etc? Pulling data from the market almost seems to be a more practical way of obtaining this information than attempting to derive it from a manual process.

[Here's a decent primer on the Oracle problem](https://blog.chain.link/what-is-the-blockchain-oracle-problem/). Not advocating that particular implementation.

cvolkernick:
Thank you @abhay for the background info will give it a read

cvolkernick:
Regarding the Sybil resistance point:

https://www.proofofhumanity.id/
https://github.com/Proof-Of-Humanity

Could potentially have other applications / implications elsewhere beyond the scope of this particular HIP, as well.

PR #273, opened 2021-09-09
HIP draft – Equitable Transmission Scaling

Tylerlhess:
An improvement to HIP-17 to encourage movement towards optimal coverage.

### DIFF
--- a/0038-Equitable-Transmission-Scaling.md
+++ b/0038-Equitable-Transmission-Scaling.md
@@ -37,15 +37,6 @@ This does add a higher degree of difficulty calculating each scaling multiplier.
 
 Rationale and Alternatives
 The main complaint is that it is possible to have a lone wolf that is transmission scaled by 30-50% due to a natural barrier such as a mountain.
-It is also possible for a single hex to theoretically create scaling to each resolution they belong to. 
-If a single hex activated 1000 hotspots we would see the following impact to those around them.
-
-res 4 | 250 | 1000 | 0.25
-res 5 | 100 | 1000 | 0.025
-res 6 |  25 | 1000 | 0.0005
-res 7 |   5 | 1000 | 0.00000025
-
-While this example is a dramatization, it proves the point that everyone inside a resolution can be greatly affected by a single res 10 hex.
 This typically happens in densly populated areas surrounded by rural areas that have no chance of increasing the resolution 4 maximum. 
 These densly populated areas may not even subjected to large amounts of scaling beyond resoultion 8 or 9 but are affecting hotspots 50 miles away.
 The sheer size and arbitray boundaries of res 4 make scaling based on it untenable. 
--- a/0038-Equitable-Transmission-Scaling.md
+++ b/0038-Equitable-Transmission-Scaling.md
@@ -7,7 +7,7 @@ Summary
 Equitable Transmission Scaling. 
 
 Motivation
-While HIP-17 made vast improvements to the equity of rewards distribution it also penalizes optimal coverage in suburban areas.
+While HIP-17 made vast improvements to the equity of rewards distribution it also penalizes movement twords optimal coverage in suburban areas.
 
 Stakeholders
 Everyone would be effected by this HIP but would benefit those at the currently scaled resolution 4 boundaries the most. This benefit would be felt
@@ -21,16 +21,16 @@ Having an equitable scaling solution that scales both directions at each resolut
 each hex down scaling to only affect the most agregious sections inside that hex while allowing optimal distributions to remain or even regain some of their downscaling. 
 This requires the scaling to be calculated form largest resoultion to smallest though.
 
-The new scaling would be the smaller of the two values of 1 or the parents scaling multiplied by the unclipped hex density divided by the clipped hex density 
+The new scaling would be the smaller of the two values of 1 or the parents scaling multiplied by the hex density divided by the target hex density 
 
-scaling = MIN( parent_scaling() * (unclipped / clipped) , 1)
+scaling = MIN( parent_scaling() * ( density / target density) , 1)
 
-res   | clipped | unclipped | scaling
-res 4 |     250 |       375 | 0.66
-res 5 |     100 |       150 | 0.44
-res 6 |      25 |        20 | 0.55
-res 7 |       5 |         3 | 0.6875
-res 8 |       1 |         1 | 0.6875
+res   | target | density | scaling
+res 4 |    250 |     375 | 0.66
+res 5 |    100 |     150 | 0.44
+res 6 |     25 |      20 | 0.55
+res 7 |      5 |       3 | 0.6875
+res 8 |      1 |       1 | 0.6875
 
 Drawbacks
 This does add a higher degree of difficulty calculating each scaling multiplier.
--- a/0038.md
+++ b/0038.md
@@ -1,0 +1,96 @@
+Author(s): BadGuyTy
+Start Date:
+Category: PoC scaling
+Original HIP PR:
+Tracking Issue:
+Summary
+Equitable Transmission Scaling. 
+
+Motivation
+While HIP-17 made vast improvements to the equity of rewards distribution it also penalizes optimal coverage in suburban areas.
+
+Stakeholders
+Everyone would be effected by this HIP but would benefit those at the currently scaled resolution 4 boundaries the most. This benefit would be felt
+on both sides of the boundaries as those outside the boundaries are no longer subject to the scaling those in their reach are not responsible for causing.
+
+How are we soliciting feedback on this HIP from these stakeholders? Note that they may not be watching the HIPs repository or even aren't directly active in the Helium Community Slack channels.
+
+Detailed Explanation
+
+Having an equitable scaling solution that scales both directions at each resolution while keeping the same resolution will cause 
+each hex down scaling to only affect the most agregious sections inside that hex while allowing optimal distributions to remain or even regain some of their downscaling. 
+This requires the scaling to be calculated form largest resoultion to smallest though.
+
+The new scaling would be the smaller of the two values of 1 or the parents scaling multiplied by the unclipped hex density divided by the clipped hex density 
+
+scaling = MIN( parent_scaling() * (unclipped / clipped) , 1)
+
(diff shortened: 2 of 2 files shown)


cvolkernick:
This definitely raises a great point about different density rules for different environments; for example, you will naturally need more hotspots in an urban environment than in a suburban or rural one, just because of all of the RF barriers an urban environment entails.

Also, if you look at Portland, Oregon for example, almost the entirety of the urban sprawl is scaled down. With areas like this, it's a situation of extremes where you either have completely oversaturated urban areas or more suburban / rural areas that have little to no relative incentive to expand coverage, due to the current universality of PoC scaling rules.

wolfenhawke:
Yes. I was looking at this just last week. In Toronto the metro area has created a saturation in Hex4, but there are plenty of areas like in Richmond Hill area (Major Mackenzie Dr/Young St) where the reward scale for res5+ is 1.00, but res4 is at 0.35. In  so the reward scale for a local HS (say Able Jetblack Sheep) is 0.35 for HIP17. So plenty of expanding urban sprawl areas are very under covered.
A way for res4 to not have such a great effect on sparse areas is desirable.

Tylerlhess:
The thought here was that by scaling up and down we could reach the target faster and closer. Especially in smaller hexes. Under this HIP an unpopulated res 7 hex would grant the first hotspot there a 1.0 scaling from as little as 0.2 scaling. which I think would be more desirable for optimal coverage than starting that HS at 0.2 there is no incentive to place there.

philltran:
[Link](https://github.com/Tylerlhess/HIP/blob/cd7d6f2afae72510c4a047a1d32644851f07bcd7/0038-Equitable-Transmission-Scaling.md) to rendered view at https://github.com/Tylerlhess/HIP/blob/cd7d6f2afae72510c4a047a1d32644851f07bcd7/0038-Equitable-Transmission-Scaling.md

cvolkernick:
Is there perhaps any way to procedurally determine the average / mean reach of hotspots on the network as a whole, and use that as a metric to determine roughly how far out (zoom out, meaning higher (lower..?) res scaling should take effect? I figure there is some version of this which currently determines the validity of PoC witnessing, and there are asserted locations that determine how far away one is from another. Just spitballing here.

Tylerlhess:
We could just drop the hex 4 scaling but I think that would reduce the opportunity for down scaling major cities correctly. I went back and forth on that and I do feel the resolutions are correct just that it shouldn't all be down scaling. There needs to be some mechanism to incentivize getting to target in each region equally. Otherwise you get a ripple effect on optimal hot spot placement instead of a gradual scaling as you near population centers.

wolfenhawke:
@cvolkernick This is a very interesting concept. The way I interpret your comment is that the "res" is dynamically created based on reach of hotspots. So, the reach of hotspots in an area may make a res10 area (as an example) based on reach of lower range antennas. So you don't really have a local hotspot effecting a hex4 area. An interesting corollary of this, is if your location is oversaturated by other hotspots, of say low gain antennas, you can increase your range and get into an area that is less covered and improve scaling. This helps the host and the network. So, the scaling is really by how many are covering an area of interest, and not by a fixed mapped shape (hex).

Tylerlhess:
@wolfenhawke the issue with this is that only a fraction of your rewards come from yourself. Its not your scaling you should be worried about but the scaling of everyone around you that you can witness.

cvolkernick:
For what it's worth, in general I'm trying to start from the first principles question(s), i.e. "what does the network want to optimize for", and then work backwards from there to think through what would the proper incentives be to encourage the behavior that leads to those optimizations.

As that applies here, my personal take is that the network wants to optimize for:

1. Coverage area -- straightforward; the total geographic reach
2. Reliability/stability/consistency of coverage -- i.e. coverage redundancies in a given area so that if one hotspot doesn't happen to hear a given broadcast signal, there are always sufficient others in the area such that at least 1 of them will receive signal.

That said, maybe what is needed is not a change to the existing scaling but an additional "layer" on top. For example, using a hotspot's witness list, calculate the average distance from the hotspot in question to all of it's witnesses. Aggregate those figures for each hotspot in a given hex, and determine a hex res average. Not sure *specifically* how it would be used (there are probably a lot of different ways one could imagine, with varying levels of technical complexity), but I'm thinking perhaps you could then use those hex averages to determine a coverage radius of each hex and somehow factor that radius into the scaling.

Essentially you would want to minimize the overlap while not leaving gaps in between. One would imagine that generally speaking urban hotspots are going to have a shorter average range on the whole, because of various previously mentioned environmental factors like buildings etc., as opposed to more suburban/rural areas with better line of sight. The areas with less reach you would want to have higher scaling than those with further reach, since *because* of the lower average coverage range you will need more hotspots to cover the same amount of area (and there generally will need to be more redundancy because of environmental blockers to "filling in the gaps" to to speak, as with large buildings which directionally obstruct said locations.

wolfenhawke:
@cvolkernick There is much value in such a direction. I actually discussed something similar for a different reason with a member some months ago. Humor me for a second. Being in the semi industry and having worked with radio chips one issue I have with them is that the RSSI and SNR measured values from a silicon chip are pretty crappy. Their error is high. This can be alleviated by calibrating every chip on the line and then doing field calibrations regularly. The manufacturing costs would go up, so it's not done on the line. The product cost would go up so it's not done in the product (field support). So, I get frustrated when I see the witness table. BUT, having a list of the hotspot's witness and their RSSI and SNR range would allow for a map of valid ranges that are expected from neighboring hotspots. Thus we are using relative values which the chip is good for, and not absolute values which the chip is bad for. This allows for a more accurate way of checking for gamers (if a new HS got added, then only the legit ones would see it quickly and be able to accurately triangulate location (maybe a backend function) and roughly check values for it.
So, I like your idea of using such a system for maybe a backend calculation of what the coverage is by some definition of an area (let's say a high hex) also.
It makes the scaling distributed. Wow. The population (of hotspots) can determine based on a calculation provided if their current reach is not providing value. The user can then not just see invalids, but be able to do something about it - longer range antenna that doesn't catch the local area or shorter range to better cover near the hotspot and not oversaturate the neighboring area.

cvolkernick:
@wolfenhawke ironic timing this was just announced by the Console team: https://blog.helium.com/coverage-arrives-with-the-latest-release-of-console-46f5858b1a93

anthonyra:
A current issue with HIP17 is the cost at which to calculate these transmit scales across the network every 30 minutes (an epoch) add another calculation on top that and it be even slower.. I like the thought of it and honestly I think an even simpler approach would be more ideal.

Even if you broke HIP17 down to it's barebones it requires a lot of calculations. As an example, let's say you wanted to calculate the reward scale of one hotspot you'd need to calculate and transverse over 49 different hexes with increasing areas and hotspot counts.

As it currently sits there's a hardcoded limit on witnesses for 300m. When comparing this to resolution 8 the max number of hotspots that can be placed simply based on this limitation is 7. One in each corner and one in the center. This is the current "max" density in the eyes of the network. If we would have a mechanism to validate each resolution 8 hex individually to the new "max" we deem necessary based on the most conservative environment we would have a more scalable approach for calculation and you as a user would only need to worry about your placement within the resolution 8 hex.

Doing it this way would also allow us to move miner rewards from the epoch to the block level which once again will move us closer to the scale needed to handle 3 million hotspots..

I too have a draft HIP that not only tries to simplify the HIP17 mechanics but also adds in anti-gaming measures and I'm not opposed to making HIP17 better but fundamentally there's already a computational problem that needs to be solved also at the very least.

Tylerlhess:
> A current issue with HIP17 is the cost at which to calculate these transmit scales across the network every 30 minutes (an epoch) add another calculation on top that and it be even slower.. I like the thought of it and honestly I think an even simpler approach would be more ideal.

This would be the same number of calculations just applying them in reverse order. Instead of closest to farthest it would flip total area of the 4 then the 5 then the 6, 7, 8, 9 and apply all the scaling all the way down instead of calculating it up. if you are calculating it up you would need to run these calculations the same number of times unless you are calculating it hex by hex. With this you could calculate state and only have to recalculate down the affected hexes with new nodes. And then any neighboring nodes that are affected by scaling up. of nodes reaching targets. It shouldn't be hard to have a relational graph and update that on the fly as new nodes are activated or old ones deactivated.

anthonyra:
> > A current issue with HIP17 is the cost at which to calculate these transmit scales across the network every 30 minutes (an epoch) add another calculation on top that and it be even slower.. I like the thought of it and honestly I think an even simpler approach would be more ideal.
> 
> This would be the same number of calculations just applying them in reverse order. Instead of closest to farthest it would flip. Calculate the total area of the 4 then the 5 then the 6, 7, 8, 9 and apply all the scaling all the way down instead of calculating it up. if you are calculating it up you would need to run these calculations the same number of times unless you are calculating it hex by hex. With this you could calculate state and only have to recalculate down the affected hexes with new nodes. And then any neighboring nodes that are affected by scaling up. of nodes reaching targets. It shouldn't be hard to have a relational graph and update that on the fly as new nodes are activated or old ones deactivated.

It needs to be performed in a decentralized manner, using a relational graph might not be so easy. There's also no need to go up to resolution 4 if you make sure that the smallest unit provides the most ideal coverage is all I'm saying.. Also I'm not saying the number of hexes being used to calculate will change but adding another calculation in general will add to the current problem and not lessen it

cvolkernick:
closed, @Tylerlhess ...?

jamiew:
@Tylerlhess was this closed by mistake? I was planning to merge this morning for further discussion. Let me know if you are withdrawing it (but lgtm)

cvolkernick:
@Tylerlhess if you are indeed closing can I have permission to copy and reopen to take up the baton? 😃 I think this is an important discussion either way. :)

Tylerlhess:
Oh the premise upon which this was opened was flawed and based on an incorrect understanding of HIP-17 clipping. I currently believe, with a better understanding of HIP-17, all that is needed might be to change some of HIP-17 N values. I will write up another HIP if and when we (in discord HIP proposal  channel sub topic HEX limit discussion) come to definitive conclusion and have better numbers to support actual impact of changing the N values compared to current scaling. Possibly dropping of Res 4 scaling and increasing res 5 and 6 N to a factor of N = 2

PR #278, opened 2021-09-17
Distributor application

chinsang26:
We at “ SANGHVI INDUSTRIES” request approval for distribution of Helium Hotspots in India  manufactured by Clodpi Labs India Private Limited and other manufactures from across the globe.

PR #279, opened 2021-09-18
Restore HIP19 document

friederbluemle:
PR #278 **accidentally deleted** `0019-third-party-manufacturers.md`. There is now a broken link from https://www.helium.com/mine (for example).

This PR restores the document that was deleted with #278.


abhay:
Good catch. I'm going to do this in a slightly different way in a separate PR.

friederbluemle:
Np :) Next time you can also ask and I can update the branch :+1:

abhay:
Yeah it was a weird request. I just wanted to have the explicit revert on the commit history.

PR #280, opened 2021-09-19
Restore hip19

abhay:
Slightly different set of commits to be an explicit revert of the last merge and an explicit re-addition. Alternative to #279. Thanks @friederbluemle

@Sophi heads up.


abhay:
@jamiew do these belong in the https://github.com/dewi-alliance/hotspot-manufacturers repo now?

PR #287, opened 2021-09-30
Add back template

PaulVMo:
It seems that the template file was deleted by a recent new HIP.


jamiew:
Ah thank you, must've slipped in

PR #289, opened 2021-10-04
HIP-42: Beacon/Witness Ratio - Witness Reward Limit

anthonyra:


anthonyra:
I'm still working on the Metabase Dashboard that will accompany this HIP I'll be sure to add it when it's done

PR #297, opened 2021-10-22
HIP draft: Standardized Firmware

flixxx:
Standardized firmware request

Rendered view: https://github.com/helium/HIP/blob/36e74635e587133a607e95634a4381496f4d85b1/0041-standardized-firmware.md


jamiew:
I agree this is an interesting idea, but I think this HIP requires more detail as far as what tha stack looks like, how control could be handed over to other parties, etc. @shawaj's comments here are a great jumping off point

shawaj:
@flixxx any further thoughts on this?

flixxx:
@shawaj I am open to whatever is the path of least resistance. I am, unfortunately, just an ideas guy and would not know where to start on execution.

During the Capcom AMA, when this was brought up - the answer was that these devices expect to have a finite lifetime. I agree in principle but usually devices die out over the course of a time period. 
If no action is taken, a large part of the network can be compromised as miners can go offline literally overnight. (e.g.: mandatory firmware upgrade after a maker exited the project).

timcooijmans:
Reflecting my comments I gave on this HIP in the DeWi community call:

I don't think firmware should be required to be standardized. I understand that there are challenges diagnostics and support tooling from some manufacturers and that there are concerns related to long term support of manufacturers. I think that the support tooling (and firmware reliability) is one of the differentiators a manufacturer can make to get a competitive edge. Nowadays a lot of manufacturers use Pi's for their platform, making it very easy to standardize firmware. However going forward with Light Hotspots we will see a lot of different platforms with a lot of different CPU architectures and even operating-systems. Some manufacturers are already using CPUs that have a development-kit under NDA. 
Additionally making a third-party firmware installation easy, also means that making a cheating firmware installation becomes very easy. With HIP19 a large part of the trust is in the firmware developed by the manufacturers. Making it easy to install third-party software and get (root) access on a hotspot means that it's very easy to cheat. 

However as soon as the trust part of HIP19 is moved away (for example with HIP22, Secure Concentrators), I'm totally ok with a community effort to develop a standardized alternative firmware for hotspots.

shawaj:
@timcooijmans it's not exactly hard for a non standardised firmware to run on a device currently even if the device in question is closed source or requires an NDA. Security by obscurity does not work and never will. Pretending that it will is a bad idea.

I think maybe this HIP should be updated though to ensure that if not standardised firmware then DeWi should have access to all firmware, even if under an NDA, so if a manufacturer goes bust the development can potentially continue.

Also the way I understand "firmware" in this context is the core code to allow the device to run on the network. This does not necessarily include support and diagnostics tooling. With light hotspots, this could essentially just be the gateway-rs build for example...

timcooijmans:
Firmware goes beyond gateway-rs. You can't just keep gateway-rs up to date because more changes are and will be needed to keep a device secure and up to date. You need everything from gateway-rs, lora-packet-forwarder to the kernel. And yes security by obscurity is currently used to deter cheaters from using the security-ICs on Helium hotspots. And yes that's not the best solution and that's why we are trying to improve as a community. 

For a big share of the SoCs out there, it's impossible to run fully open-source firmware because many of the "development-kits" provided by the vendors that are required to compile firmware/Linux are non-public, require a license and/or NDA. SoC vendors don't do this because of platform security but to protect their IP. You can't just compile a Linux kernel for those SoCs because you are missing device-drivers that are essential. And if some level of support is available in the open-source eco-system you are regularly missing important drivers, resulting bad performance and missing features.

I don't see how giving DeWi access to the source-code is going to work:

- DeWi doesn't want to sign NDA's, and I can understand that because it's a huge liability especially when dealing with IP from big silicon vendors. 
- An NDA doesn't always give DeWi the permission to use the development-kit and to create firmware images with it for the general public. The silicon vendor would have to be involved in signing an agreement. 
- Additionally there's the issue of secure-boot that some vendors in the Helium ecosystem are already using (because using TrustZone kind of dictates that you at least do some level of secure-boot). Secure-boot signing keys also need to be transferred to DeWi in this case. 
- Who's going to do the development for hotspots from a manufacturer that went bust? DeWi can't just share the source-code with a third-party or open-source with the community it because it's under NDA? It can be a multi-person full-time job to maintain firmware images. I don't see how this is going to work. 

Maybe if you come from a world of using DIY-like boards like Raspberry Pi's, Rockchip Pi's or alike that this seems easy. But if Helium want's to professionalize and scale, it needs to move from the "slap a concentrator on a Pi"-phase to getting support on specifically designed LoRa gateways that use SoCs from big vendors that can deliver in high volumes. 

So maybe @flixxx can comment on his intentions of this HIP? What do you mean by firmware @flixxx? Is it:

1. The full image including the OS (Linux Kernel most probably) and everything required to run the hotspot on the Helium network (excluding custom support and diagnostics tooling) and tools required to create full image.
2. The Helium specific code, so gateway-rs. if modified. In this case, how should updating work if a vendor goes bust?

My proposal would be to make compatibility with a standardized firmware optional. In other words as vendor you can sign up for the "hotspot-rescue-program" (I don't see why DeWi should run this program, can be a separate organization or maybe just publishing code to Github or elsewhere) and include this in your marketing so buyers can decide if they take this into account when selecting a vendor.

mfalkvidd:
> @Security by obscurity does not work and never will. Pretending that it will is a bad idea.

This is unfortuntately a common misunderstanding. There are lots of references on this misconception. For example, wikipedia [says](https://en.wikipedia.org/wiki/Security_through_obscurity#Criticism):

> When used as an independent layer, obscurity is considered a valid security tool.

> NIST's cyber resiliency framework, 800-160 Volume 2, recommends the usage of security through obscurity as a complementary part of a resilient and secure computing environment.

Thedailygrind:
@shawaj Batch 3 indoor UK still not arrived and customer service saying will be delivered in the New Year. You must be genuinely embarrassed at Nebras incompetency?

shawaj:
@timcooijmans 

> You need everything from gateway-rs, lora-packet-forwarder to the kernel

All of which are open source and therefore I don't see any reason why they shouldn't be enforced to be by DeWi

> Maybe if you come from a world of using DIY-like boards like Raspberry Pi's, Rockchip Pi's or alike that this seems easy. But if Helium want's to professionalize and scale, it needs to move from the "slap a LoRa concentrator on a Pi"-phase to getting support on specifically designed LoRa gateways that use SoCs from big vendors that can deliver in high volumes.

Haha. This makes me laugh. The biggest vendor on the network is using a Rockchip CPU which is available in much higher volume than anything else currently available... Rockchip is one of the biggest ARM CPU vendors in the world. And the next highest occurrence of boards is Raspberry Pi boards, which as your well know is used in production by all kinds of companies. You trying to belittle people using the most successful SBC platforms in production is kind of ridiculous to be honest. And "professional" in this context just seems to me to be some kind of hardware snobbery rather than being based on any actual technical merit.

Making standardized firmware optional is a pointless endeavour. The vast majority of vendors (as per now) will choose not to and because the market is constrained right now they will get away with it. Additionally, a lot of the people participating in the network do not have the technical knowledge to even understand the implications of this and so it is on DeWi to provide these kind of assurances IMO.

There is no reason why people participating in the Helium community should not open source their software (and hardware) since they are building on open source software and hardware themselves.

shawaj:
@mfalkvidd show me an OPSEC professional that says that obscurity by itself is a reliable way to provide security and I'll happily eat my hat! 🤣

timcooijmans:
> All of which are open source and therefore I don't see any reason why they shouldn't be enforced to be by DeWi

gateway-rs and lora-packet-forwarder are licensed such that modifications don't need to be open-sourced. So while they are open-source it doesn't mean that all derivatives should be open-sourced, it's a good gesture but there's no requirement to do so. Additionally as I discussed before, Linux-based firmware for SoCs commonly contain a number of linked proprietary hardware drivers that cannot be open-sourced. These drivers are provided as linked modules by the software development kits provided by SoC manufacturers under NDA or similar licensing. 

>There is no reason why people participating in the Helium community should not open source their software (and hardware) since they are building on open source software and hardware themselves.

The Helium community and Helium Inc is open-sourcing their main software because it's required by all parties to run the Helium network and blockchain and is to be trusted (remember the software _is_ the blockchain). They want it to be a community project. 
But just because we use a blockchain network that used open-source code doesn't mean that we should open-source every single component that is required to run the Helium network on our hotspots. For example the firmware/operating-system on hotspots is not part of the Helium network and blockchain. 
A lot of banks, just as an example, are using Linux and open-source software libraries and frameworks for their payment systems and core banking systems. By your analogy they should also open-source all their core banking system code because they are building on open-source software?  
As I explained before, there are reasons why it's not possible to open-source _everything_ that's required to build the firmware that an hotspot uses.
Additionally even Helium Inc's firmware (and/or the tools to re-create that firmware) for the OG hotspots and RAKs is not open-source. So apparently Helium Inc doesn't consider the firmware part of the open-source stack. 

I'm not sure where you got the idea that Helium's hardware is open-source. Sure you at Nebra voluntarily open-sourced some hardware designs. But the big majority of the hardware (again including Helium Inc's OG hotspots) is not open-source. The Helium GitHub contains 0 hardware repositories.

cvolkernick:
> @shawaj Batch 3 indoor UK still not arrived and customer service saying will be delivered in the New Year. You must be genuinely embarrassed at Nebras incompetency?

How many did you order I have some unopened from batch 3 I will ship you them if it will stop all this. C'mon man this back and forth has gone on literally months now enough public shaming Shawaj please, I'm sure he already deals with enough having all of these problems.

Thedailygrind:
I see it as all extremely relevant. The fact he still has a know it all attitude running a company with such major failings should be even more concerning. 

I have 3 on order. If you'd do that @cvolkernick then I'll gladly accept. I'll send you my order confirmation as proof if you provide contact details. 

I appreciate what you're saying but @shawaj public shame is literally nothing compared to what some people have had to go through because of Nebra.

cvolkernick:
I'm not attempting to invalidate your difficulties, only stating the fact that the horse has been beaten to a pulp at this point. It's clear DeWi & Helium Inc are recusing themselves from any judgement on this matter in particular, so continuing down this GitHub excoriation path doesn't seem productive. Literally been there my friend, it sucks but we have to process & move on now.

And again this is not a HIP-19 or Nebra-related PR -- this is about Standardized Firmware.

Thedailygrind:
How you gonna have standardised firmeware without standardised manufacturing capabilities? @cvolkernick

jamiew:
This HIP is still lacking the necessary detail for a productive conversation around implementation. @flixxx @shawaj @timcooijmans do any of you have any interest in taking it further? Otherwise we will close as inactive. 

Please note it can be re-opened at any time! Thank you

shawaj:
@jamiew i think it is pretty obvious we should take this further. without it, once some hotspot manufacturers inevitably collapse, large numbers of customers will be left with some expensive doorstops. Not good for the environment, not good for customers, and not good for the network.

if this is not done, as an alternative, DeWi should enforce taking some more technical data on the chipsets and various other pertinent data that would allow DeWi and/or another manufacturer or helium community member to take on the software updates and support for a failed vendor.

all of @timcooijmans issues with this seem to be to do with licensing issues - none of which are reasonable or valid, IMHO. For example Raspberry Pi maintains proprietary firmware blobs, however they still manage to have an open source repo of their Linux kernel and software stack. There is no reason why this could not be done for all manufacturers.

My analogy has nothing to do with banks. Banks are not decentralised, open networks by design. @timcooijmans if you are suggesting that Helium Network should become like a bank then I worry you are part of the problem not the solution here.

Re your talk about hardware not being open sourced - this HIP is solely about firmware/software. I don't see how the hardware comes into it. My mention of hardware was more just to make a point, rather than a suggestion we should change the HIP to include hardware. However, the "block diagram" and various other pertinent things to the firmware updates (reset pin GPIOs, datasheet etc) should be, at the very least, communicated privately to the DeWi.

Just to reiterate - this HIP is about standardised firmware. That does not, in my opinion, extend as far as what OS you have to use or anything to that effect, or to the hardware. The purpose here is so that there is a reasonable path for customers to still receive updates from other sources should the manufacturer in question not be providing them (for whatever reason)

jamiew:
Sure, I am happy to re-open. Given the kind of technical information you mention, do you think it is realistic that a another company would assume responsiblity for another vendor's abandoned hotspots? e.g. what incentive do they have to do that extra work?

jamiew:
Hi, it's been a few months, I'm marking this stale. @shawaj did you still want to pick it back up? Otherwise I'll close again until folks have time. Thanks!

shawaj:
Yeah keep it closed for now. What I'll do is create a new PR as I don't think "standardised firmware" is the objective now and we are applying for a DeWi grant as well so it should match to that.

PR #324, opened 2021-12-20
Taiwan Shengwei Electronic Technology Co., Ltd.md

Eroskwokgithub:


Eroskwokgithub:
We have re-studied the course and submitted the manufacturer's application here, but it is not clear whether the application is correct. If there is any mistake, can you tell us and guide us to operate correctly? thank you

jamiew:
@Eroskwokgithub the correct place for mgr applications is https://github.com/dewi-alliance/hotspot-manufacturers

Eroskwokgithub:
Thanks for your tips and help. We will resubmit when we are done! Thanks
again

Jamie Dubs ***@***.***> 于2021年12月29日周三 10:41写道：

PR #329, opened 2022-01-14
HIP-42: grammar and clarity improvements

dogweather:
HIPs should be judged on their merits. To help with this, grammar, spelling, and style should be smoothed out where ever possible. 

The HIP audience is generally Hotspot owners who may not have read the whitepapers. I believe it's important to tailor a document to its audience.

### DIFF
--- a/0042-beacon-witness-ratio-witness-reward-limit.md
+++ b/0042-beacon-witness-ratio-witness-reward-limit.md
@@ -11,7 +11,10 @@
 # Motivation
 [motivation]: #motivation
 
-To ensure that hotspot owners are motivated to not just prioritize witnessing but also to beacon.
+Twofold: 
+
+1. To ensure that hotspot owners are motivated to build a fully working IoT network. Both witnessing and beaconing are essential.
+2. To combat recent gaming that has been observed.
 
 # Summary
 [summary]: #summary
--- a/0042-beacon-witness-ratio-witness-reward-limit.md
+++ b/0042-beacon-witness-ratio-witness-reward-limit.md
@@ -16,7 +16,7 @@ To ensure that hotspot owners are motivated to not just prioritize witnessing bu
 # Summary
 [summary]: #summary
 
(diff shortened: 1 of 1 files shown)


dogweather:
@anthonyra "determine coverage without the need for devices"

I don't understand what that means. Hotspots are devices. Do you mean, devices apart from the hotspots?

dogweather:
@anthonyra I added more context to the **Motivations** section based on confusion I've witnessed.

anthonyra:
> @anthonyra "determine coverage without the need for devices"
> 
> I don't understand what that means. Hotspots are devices. Do you mean, devices apart from the hotspots?

Ahh, there could be added clarity here. I'm in a way combining the roots of what Proof of Coverage intentions are along with what this HIP will improve upon.

anthonyra:
> All hotspots that earn witnessing rewards while beaconing far less than the network average of 2.4 beacons/day.

Well, kind of. Yes as of right now the way PoC is incentivized witnesses are indeed rewarded more than beacons. This however is leading to an imbalance when it comes to people shooting for "optimal" hotspots solely based on witnesses. It's easier to find a location that has the potential to witness a lot of hotspots and then factor in Man In The Middle attacks that can lead to witness repeating you see concerning network behavior. This however, isn't a surprise based on the current network. It's incentivized more to witness plus you have more control to "affect" witnesses then beacons. This is trying to link beacons to witnesses but in a way that doesn't 100% put people up to chance with beacons.

The alternative to the first motive for this HIP is simply to change the Reward Buckets. But if you do that people will be 100% at the mercy of the network and being beaconed or not.

I also have a list as of yesterday of all affected hotspots. We could add this as more data to see affects of this HIP but it'll need some very clear notes and disclaimers...

anthonyra:
It'll probably be a great idea to have a huge disclaimer at the top describing what it means to be a Temperature Check vote since that's something new with this HIP but is expected to be standardized with future HIPs.

A Temperature Check, is a vote to see if the community approves of the concept or improvement being suggested by the HIP. At the core, if there's network issues or concerns that might prevent the HIP to operate in the best interest of the network there either needs to be a suggested means to reduce the impact or the implementation must be delayed until those concerns are fully addressed. This vote, also green lights the process of converting this HIP to code to include Core Developer mentorship and review. Once code is complete and reviewed the Final Check vote, (*name to be determined*) will be conducted to secure implementation window and final chain variable settings.

dogweather:
I definitely think the URL of the dewi check tool would be great to place up high in the doc. It's a frequently asked question, and can dispell a lot of worry.

dogweather:
> Yes as of right now the way PoC is incentivized witnesses are indeed rewarded more than beacons.

Will this HIP change that?

jamiew:
@dogweather is this OK to merge? changes look Ok to me, always appreciate copy editing. @anthonyra lmk if you disagree

anthonyra (approved):
Looks good to me

Issue #336, opened 2022-01-27
HIP 51: Helium DAO


hiptron:
- Authors: [@tjain-mcc](https://github.com/tjain-mcc), [@shayons297](https://github.com/shayons297), [@jmfayal](https://github.com/jmfayal),[@abhay](https://github.com/abhay)
- Start Date: 2022-01-26
- Category: Economic / Technical
- Original HIP PR: https://github.com/helium/HIP/pull/334
- Tracking Issue: https://github.com/helium/HIP/issues/336 (this)
- Status: Approved

# Rendered view

Read the HIP:

https://github.com/helium/HIP/blob/master/0051-helium-dao.md

# Summary
[summary]: #summary

This proposal outlines economic and technical constructions with the aim of scaling the Helium Network to support new users, devices, and types of wireless network protocols.

We propose that each wireless network supported by Helium (LoRaWAN, Wifi, 5G — referred to as wireless network protocols or WNPs) has its own subDAO with its own token (referred to as wireless network tokens or WNTs). The key specifications of the WNT such as proof-of-coverage rules, data credits rewards, and consensus mechanisms are governed by each WNT DAO.

The aim is to create an economy such that the underlying HNT-Data Credit burn-and-mint equilibrium remains undisturbed, and Proof-of-Coverage rules and miner emission amounts are dictated by the corresponding wireless network’s sub-token.

The technical model requires that the entire Helium Network lives at some proposed base layer (L1). All accounts and transaction activity happens at this L1. All subDAOs operate as economic and governance layers (L2) with their respective WNTs acting as native tokens for this purpose. Each L2 also runs software to calculate WNP specific items (eg: all the proof of coverage code that exists on the current LoRaWAN network to determine hotspot rewards) and the L2 validators who have staked WNTs come to consensus on these calculations.

ckizildag1907:
Do you think the proposed HIP could support networks other than star topology? Does a mesh network could become a WNP ? Mesh Proof of coverage is slightly different than device to hotspot architecture of star networks. (Device to devices than to hotspot)

jpvam:
These proposals are great ideas HIP 51, 52. There are some fundamentals that would build confidence in the expected target outcome:
1. Is DeWi & Helium committed to drive the maximum balanced value for both Lora and 5G/Wifi.  With recent observations in technical debt, skills and changing tokenomics, limited consideration for SLAs -- are you tooled to make this successful and industrial grade. 
[...]

PR #355, opened 2022-02-05
Update 0051-helium-dao.md

shayons297:
#336


jamiew:
commit from named author, merging

PR #363, opened 2022-02-23
changes to support validator pool as challengers

andymck:
Updates to describe the wider validator pool as challengers rather than having this role limited to consensus group members


jamiew:
PR by named author, merging automatically beep boop I'm a bot

Issue #338, opened 2022-01-27
HIP 52: LoRaWAN subDAO


hiptron:
- Authors: Tushar @tjain-mcc, Shayon
- Start Date: 2022-01-04
- Category: Economic / Technical
- Status: Draft

# Summary

This proposal includes a specification of the LoRaWAN Wireless Network Protocol as per the DAO model and L2 implementation outlined in HIP 50.
Summary

In HIP 50: Helium DAOs, we provide a general structure for onboarding new WNPs to the broader Helium Network, with mechanisms in place to ensure that protocol-specific attributes such as proof-of-coverage rules, data credits pricing, and block validation are within control of the emergent WNT DAO.

In this proposal, we specify the implementation of the structure proposed through a detailed onboarding proposal for the LoRaWAN Network. We propose initial configurations of the LoRaWAN economics layer as well as governance mechanisms within the DAO through LoRaWAN (LRW) token voting.

# Rendered view

(to come, stand by)

shawaj:
In #53 for 5G subDAO we have a mechanism for hotspot vendors to get paid for their support / updates / deployment strategies as follows:

> Hotspot vendors can adjust what % of MOBILE generated by the fleet of hotspots deployed by their customers is kicked back up to the hotspot vendors and pursue different deployment strategies, from giving out free/highly subsidized hotspots to passing 100% or rewards to hotspot operators, but charging higher $$ prices for hardware.

It seems to make sense that this functionality should be built into the network mechanics of all subDAOs, and certainly for the IOT / LoRaWAN one.

Currently, hotspot vendors are expected to provide a significant amount of infrastructure for a one off flat fee piece of hardware. Whilst this, on the surface, appears to be good for end customers, it ignores a pretty major flaw that after some period a lot of these devices could potentially become unsupported with updates due to their being no ongoing financial compensation for the infrastructure and engineering work to deliver continuous updates.

It seems very wrong to me that currently hotspot vendors, despite being one of the core members of the Helium ecosystem, are not rewarded for this activity when practically all other network participants with ongoing engineering and infrastructure costs are rewarded in some way.

This seems to be a bit of an oversight currently and it seems to me that the transition to IOT token would be a great time to fix this.

shawaj:
Another added benefit this could bring is that Helium Foundation/subDAO could perhaps have some sort of control over a mechanism for other parties to take over firmware updates for hotspots should a vendor disappear or similar. And there would be an incentive for community members or other vendors to actually take this on.

[...]

PR #394, opened 2022-04-28
4/29

shayons297:
#338


jamiew:
hi @shayons297, is this and/or #393 ready to merge? they're both still listed as Draft. I can override that but typically author will do a 'ready for review'

PR #414, opened 2022-05-24
HIP draft – PoC Witness IP Check

H-Baguette:
An attempt at solving Helium's spoofing problem. A discussion with Digerati on Discord allowed me to better understand why the simple solution to invalidate all witnesses sharing the same IP as either the beacon or another witness is not ideal.

Ed: rendered view https://github.com/helium/HIP/blob/defd2f4233650967b548e4115e4f5e4a062030ed/0061-poc-witness-ip-check.md


edakturk14:
hey @H-Baguette! would you like to present the draft on the community call thats tomorrow?

snip:
Looks to be a very good start thanks @H-Baguette!
Even if it is a good step forward it seems this proposal will not stop cheaters using VPN and unique IP to go outside.
But it is difficult to cover everything. So for sure small steps are better than doing nothing!

softlion:
You could also check that IPs are residential and exclude all other types (ie: datacenter IPs).
Explanations here: https://multilogin.com/how-to-hide-ip-address-your-complete-guide/

snip:
> You could also check that IPs are residential and exclude all other types (ie: datacenter IPs). Explanations here: https://multilogin.com/how-to-hide-ip-address-your-complete-guide/
> 
> image

Unfortunately lot of current hotspot owners moved to VPN for their hotspot to have possibility to forward port 44158 to their hotspot. So they will need to update their setup to remove the VPN.
But yes this can be a target solution. Not yet applicable i think.

ifeign:
> > You could also check that IPs are residential and exclude all other types (ie: datacenter IPs).
> 
> Unfortunately lot of current hotspot owners moved to VPN for their hotspot to have possibility to forward port [44158](tel:44158) to their hotspot. So they will need to update their setup to remove the VPN. But yes this can be a target solution. Not yet applicable i think.

People also use VPNs for remote management

softlion:
> > > You could also check that IPs are residential and exclude all other types (ie: datacenter IPs).
> > 
> > 
> > Unfortunately lot of current hotspot owners moved to VPN for their hotspot to have possibility to forward port 44158 to their hotspot. So they will need to update their setup to remove the VPN. But yes this can be a target solution. Not yet applicable i think.
> 
> People also use VPNs for remote management

For management they use a secondary network on vpn, which does not get all 'normal' traffic from hotspots.

ifeign:
I understand the gist of things, but I have access to some acreage where I
could deploy miners 300m apart but I’d probably use the same public IP. If
this HIP were to be passed would I be screwed? Would I need a VPN for each
miner so it has a unique IP?

hiptron:
This HIP draft has been numbered and merged for discussion as HIP 62. 

Please direct future questions & comments to the new tracking issue: #422 

If you are one of the named authors, please include `#422` in future pull requests to have them automatically merged.

PR #439, opened 2022-06-30
HIP draft - MCC Reconstruction

TheRealJohnMac50:
Initiating HIP 67 - MCC Reconstruction 

Ed: rendered view https://github.com/helium/HIP/blob/ac42ba268337c07820020bde01ea808e918f3dd8/0067--MCC_RECONSTRUCTION.md


jamiew:
Hi, thanks for your submission

A couple of notes on this:

- I believe it would be more useful to frame this as simply "limiting the authority of the MCC" or similar rather than focusing so much on "prevent innocent people being harmed." The facts and overall intention are sufficient. Focusing on the harm to Panther or Deeper hotspot owners ignores the harm to the entire network (and its credibility) from the presence of industrial-scale cheating
- it would be useful to reference any other materials that outline the goals, structure and membership of the MCC, if possible. readers could better understand what it currently does vs. what you're proposing to remove

jamiew:
I've also de-numbered the proposal

shawaj:
This will need to probably go hand in hand with #297 which I'm going to redraft as soon as I get some time.

We already have invested many tens of thousands in trying to build a standardised firmware and are planning to apply for a Helium Foundation grant to continue this work with Balena.

Meaning that people who are effected would still be able to get updates even if a banned maker stops providing them.

And also, as a separate HIP I think that makers should get a share of the hotspots earnings, even if it's 1 or 2%.

This can then inventivise a new community member or maker to take over the updates and onboarding of new devices from a banned maker

TheRealJohnMac50:
> I've also de-numbered the proposal since it is still draft

Please delete (if possible) this hip proposal.

jamiew:
@TheRealJohnMac50 you got it. marking as closed/withdrawn

PR #440, opened 2022-06-30
Update and rename 0000-template.md to 0067-MCC_Reconstruction.md

TheRealJohnMac50:


jamiew:
Hi, closing as a duplicate of #439 since these look functionally identical? please re-open if not

PR #473, opened 2022-09-02
Nit on HIP 70 wording

syuan100:
- Definitively saying that we will move to decentralized oracle operation
- Clarification on what kind of data will be available while we have centralized oracle operation


abhay:
Addressed in #475

syuan100:
Changes included in #475

PR #482, opened 2022-09-15
HIP draft - Reduce IoT Use Cost

oosui:


vincenzospaghetti:
Hi @oosui - thanks for submitting this HIP. I am a HIP Editor and have a few questions. Can you send me your discord so I can reach out to you with questions?

Thanks,
Valerie

vincenzospaghetti:
Hi @oosui want to make sure this gets attention. Can you follow the HIP template to fill out the necessary sections to make this HIP complete?

See template here: https://github.com/helium/HIP/blob/main/0000-template.md

wolfenhawke:
I don't buy the argument that as a business provider, you will have to field your own hotspots - that is completely contrary to the helium purpose. Instead, what we will find, is that when the network really starts to get used, then overconcentration of hotspots will alleviate the risk that is brought up. It also alleviates the risk of lost detections due to too many users. Think of a bike tracking service in NYC, and hoping one hotspot picks up all that passes it - not going to happen. But redundancy, like the friend for blockchain, will help here. If hotspot1 doesn't pick up the Tx because it is busy with the first bike that it captured, hotspot2, which may be in the same building has a good chance, or hotspot3, etc. This also covers the risk the OA brings up which is hotspot1 may be turned off.

waveform06:
Anyway the cost additional cost of a Helium data-only hotspot (No ECC) over a TTN gateway is $15. Not $60
Same gateways that TTN uses can usually be used plus a data-only onboard of $10 and a location assert of $5 (which is only needed to locate within a Region and is uncaring about the actual address, so unlikely to need a 2nd.
So basic data-only hotspot cost for Helium = 
100,000 / 1000 * (>$140+$10+$5) > $15,500 not $20.000
Plus you should need less or possibly Zero in some locations as the Helium network is already redundantly present.

Which goes down to the cost of the Sensor uplinks. And if the project cant save $1 a year per sensor maybe it isn't a good use case for a LoRaWAN deployment.

oosui:
> Hi @oosui - thanks for submitting this HIP. I am a HIP Editor and have a few questions. Can you send me your discord so I can reach out to you with questions?

Hi @vincenzospaghetti, sorry for the late reply. 
I think shouldn't have created a PR since I don't have a detailed solution for this and my knowledge on Helium Network is limited. Maybe I'm wrong. Let me just create an issue, you can edit it and if it's worth to create a PR, can you please find someone is very familiar with Helium Network? Thanks!
https://github.com/helium/HIP/issues/483

vincenzospaghetti:
Hi @oosui that sounds fine! Let's get you more integrated as well. If you're not in our Discord, please go to helium.foundation and join the community there. We also are on Reddit & Twitter and you can stay up to date with happenings that we post.

In the meantime, I'll close this HIP and you can always write a new PR in the future. Thank you!

PR #515, opened 2023-01-03
Updates HIP74 Genesis references

meowshka:


vincenzospaghetti (changes requested):
Hi Sasha - this looks great, but can you add some clear language about:
- This HIP is to be PoC post-Genesis, but we might use these tools pre-genesis ending
- We can start implementing this and treat it as beta
- clearly state there will be other HIPs proposed to communicate the changes to MOBILE PoC, and the HIPs going forward will be smaller and more easily consumable
- This HIP does not impact the Genesis Phase

Essentially it needs language that says its not impacting the premine nor is it kicking off programmatic emissions. It will take effect once the Genesis Phase is over. And, if it is implemented pre-Solana transition because the premine runs out, the core devs will work to begin programmatic emissions on the Helium L1 and treat it as beta.

PR #518, opened 2023-01-10
Solving the IOT Sensor Cold Start Problem

maxgold91:
**HIP Title: Solving the IOT Sensor Cold Start Problem**

**Author: Max Gold**

**HIP Abstract:**

This HIP proposes that 25% of the IOT rewards distributed on the IOT subDAO be allocated to sensors for their contribution to the network. Sensors play a vital role in the Helium Network by proving reliability and uptime confirmation for the IOT network and ultimately burn DC. By allocating a portion of the IOT rewards to sensors, we can incentivize the deployment and maintenance of sensors, which will ultimately benefit the entire network.

**HIP Content:**

**Motivation**

Currently, sensors do not receive any rewards for their contribution to the network. Currently very few active devices use the network and very little HNT is burned for data transfer.  This HIP proposes that 25% of the IOT rewards be allocated to sensors as a way to incentivize their deployment and maintenance.

**Rationale**

By allocating a portion of the IOT rewards to sensors, we can encourage more people to deploy and maintain sensors, which will ultimately benefit the entire network. It is likely that the financial incentive from IOT rewards for sensors will cause a gold rush effect, as more people will be motivated to deploy sensors in order to earn rewards. As the number of sensors on the network increases, the demand for LoRaWAN components will also increase. This increased demand could potentially drive down the price of these components, as manufacturers ramp up production to meet the demand. Lowering the price of LoRaWAN components would also likely result in a reduction in the price of sensors that use these components, making it more affordable for individuals and organizations to deploy sensors on the network.  In 2019 a LoRaWAN gateway cost over $1,000, in 2023 they can be purchased for $50.  We hope to see a similar reduction in price making LoRaWAN enabled sensors more financially viable compared to other options using NB-IOT.

Additionally, the fact that the Helium Network is open and accessible to all types of devices, including even the most basic sensors, means that we welcome a wide range of connections to the network. We want as many devices as possible to be able to connect to the network and benefit from the improved coverage and connectivity that a large sensor network provides. Whether it's a high-tech device or a simple pet rock, we want all types of devices to be able to connect to the network and contribute to its growth and success.  This serves another purpose as it allows us to stress test the current network.  It is much better to weed out bugs when sending test data during the initial phase.

**Implementation**

To implement this HIP, we will need to modify the reward distribution algorithm to include a percentage of the IOT rewards for sensors. The exact percentage of the rewards that should be allocated to each bucket will be determined based on feedback from the community and further analysis of the impact on the network, however for now 25% will serve as a placeholder.

**Drawbacks**

This HIP can be easily gamed, however I don't think we care.  

**Success Metrics**

We hope to see the price of LoRaWAN sensors decrease by about 50-75% within 2 years.

### DIFF
--- a/0081-minimum-onboarding-fee.md
+++ b/0081-minimum-onboarding-fee.md
@@ -8,12 +8,10 @@
 
 ## Summary
 
-This HIP proposes to set a minimum onboarding fee for Hotspots in the subDAO structure. This HIP also proposes that Hotspots for which the minimum onboarding fee has not been burned are not eligible for rewards from the subDAO. The minimum onboarding fee will be dynamically defined as $40 and will follow the same halving schedule as HNT and would theoretically start at $20.  This HIP will be implemented in phases, first clearly defining the requirements at the time of its passing, then on-chain with a target date of August 1, 2023.
+This HIP proposes to set a minimum onboarding fee for devices in the subDAO structure. This HIP also proposes that devices for which the minimum onboarding fee has not been burned are not eligible for rewards from the subDAO. The minimum onboarding fee will be dynamically defined as $10 on August 1, 2023 and will follow the same halving schedule as HNT emissions as shown in the chart below.  It is important to note, this HIP only defines a minimum fee and at the time of its writing, HIPs 51-53 have defined the onboarding fees as $40 per hotspot.  If a subDAO wishes to change its onboard fees, it must do so through the subDAO governance process.  This HIP will be implemented in phases, first clearly defining the requirements at the time of its passing, then on-chain with a target date of August 1, 2023.
 
 | Date       | Minimum Onboard Fee in DC |
 |------------|--------------------------:|
-| 08/01/2019 | 4,000,000                |
-| 04/18/2021 | 2,000,000                |
 | 08/01/2023 | 1,000,000                |
 | 08/01/2025 | 500,000                  |
 | 08/01/2027 | 250,000                  |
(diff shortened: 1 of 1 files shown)


waveform06:
How do you define ownership of a sensor to reward it? 
You don't even need a wallet on the blockchain to deploy sensors so how do you reward them?

maxgold91:
Oh so the problem there would be anyone can fund the sensor regardless of ownership thus flowing rewards to the wrong person.  Hmmm, I wonder with the move to Solana if we can easily tie an NFT to a sensor to overcome this hurdle.  Obviously I'm not the coder guy, just throwing out economic incentives to get more sensors into production which should help make LoRa sensors more financially viable.  Industry feedback at CES last week seemed to be that LoRa is both too expensive and unreliable.  There's not much we can do on the perception of reliability front, just trying to throw out ways of helping the "too expensive" problem.

waveform06:
Sensors are not present on the blockchain, remember PoC is moving off-chain and packets are already off-chain.

maxgold91:
Right but how difficult would it be to have an opt-in sensor as an NFT option to earn a percentage of the IOT token?  If the goal is to have your pet rock on the internet, cheap connectivity isn't enough.  Need to have cheap sensors too.  I'm open to other options but tokens have a history of reducing hardware costs.

abhay:
The NFT idea isn't a bad one per se. Might be hard to onboard one per device, though. 

An alternative could be a dev addr slab be the NFT and oracle verified usage on that slab (derived from helium packet router data). You have to stake to get the NFT but it derives some kind of IOT earnings that's scaled based on your usage (with some reasonable caps).

maxgold91:
Hmmm maybe rewards that are skewed towards simply existing and passing a minimum amount of data with a decay rate on reward units earned for each incremental tranche of data used.  Want the goal to be more about buying sensors, not a return to the pre-HIP 10 days.

vincenzospaghetti (approved):

PR #547, opened 2023-01-25
Create 00XX-linear-lockup-curve.md

ferebee:
HIP draft to simplify veHNT lockup curve


abhay:
@samgutentag any ideas why HIP 49 is getting updated in here?

vincenzospaghetti:
@samgutentag @abhay it looks like it might be the prettier because I can't see any text changes. 
How can we instruct the other to remove any changes to hip 49? even if its just the prettier.

ferebee:
Sorry here for the confusion. It looks like the spurious edit to 0049-max-eirp-adjustment.md was caused by the prettifier. AFAICS I have now reverted that, leaving just the revision to the actual HIP draft.

PR #554, opened 2023-02-05
Location Assertion Requirement HIP

ilovespectra:
Wrote a HIP in hopes to reduce some of the concern around the DC rewards in allegedly gaming hotspots.


vincenzospaghetti:
@ilovespectra let me know when you've made the updates we discussed in Discord.

waveform06:
There are a lot of data-only hotspots that ARE asserted that people complain they think they are cheating. eg many on this list
https://explorer.helium.com/txns/iR8AGP1BIuCBmesPERLBzNAfpeqyE-ZsQcAu19cStm8
So how would making them assert remove suspicion for these - which is a much bigger group?

As Data-Only do not participate in PoC and cannot be triangulated by fixed location transmissions they could be asserted anywhere in the region and still look valid. As is the case with most of the above.

And can you provide more explanation of "Lastly, implement a mechanism that will attempt to redirect Data Transfer to eligible [asserted] hotspots. In the event there is not a hotspot in range to handle the Data Transfer, the communication will not be completed." as I dont think that is technically possible as its described in LoRaWAN.

ilovespectra:
I hear the concern about asserted hotspots and assertion dishonesty. I’d like to brainstorm ideas to help encourage honest participation. Realizing this HIP will not force it anymore than existing guidelines.Regarding the implementation of a re-selecting process, I suppose I imagined a mechanism similar to how beacons seek 14 witnesses and provide invalids when necessary. I imagined something along the lines of a try catch block that simply checks for assertion. In the event it is not asserted I don’t know exactly how the device would forward the data from the denied packet via RF like a beacon to other hotspots, but imagine there would be a way to encrypt and send that over radio after the DO hotspot receives the packet and is unable to forward it directly to the network. This would require technical know how outside my pay grade but I’d be happy to rework this phrasing and implementation alongside any qualified individuals in order to adequately address these issues. I really appreciate the feedback!On Feb 16, 2023, at 2:44 AM, Adrian ***@***.***> wrote:﻿
There are a lot of data-only hotspots that ARE asserted that people complain they think they are cheating. eg many on this list
https://explorer.helium.com/txns/iR8AGP1BIuCBmesPERLBzNAfpeqyE-ZsQcAu19cStm8
So how would making them assert remove suspicion for these - which is a much bigger group?
As Data-Only do not participate in PoC and cannot be triangulated by fixed location transmissions they could be asserted anywhere in the region and still look valid. As is the case with most of the above.
And can you provide more explanation of "Lastly, implement a mechanism that will attempt to redirect Data Transfer to eligible [asserted] hotspots. In the event there is not a hotspot in range to handle the Data Transfer, the communication will not be completed." as I dont think that is technically possible as its described in LoRaWAN.

waveform06:
Some background info for you.
A data-only or non PoC capable DIY hotspot does not need to be asserted to receive sensor data (uplinks). But it does need to be asserted somewhere in the correct region to send downlinks to sensors. As most sensor deployers probably need to send downlinks to sensors it could be assumed that new data-only hotspots that dont assert are not related to real sensor deployments. 
There is no requirement for a data-only to be accurately located and it is not possible for this to be validated.
There is a theoretical limit to how many DC a real hotspot could pass in a day of real sensor traffic. Some hotspots transfer more than this so the sensor data is almost certainly all faked.
As for sensor data working like PoC, well currently all hotspots do pickup sensor data and its sent from all hotspots to Helium so there is no need for hotspots to start forwarding data, as there is a decision at the back end who to pay out of the hotspots that send data.
On from that ...If you have a hotspot that is based in a sealed room or even if its a virtual hotspot with a fake location and its passing fake data why would we forward this on for payment by another hotspot? 
Or If the location were unasserted and it was in a real place and the all of the data was real (but forced) and not picked up by other real hotspots  I dont know how you could decide which other hotspots to pass the data thru as non of them picked up the signal and we dont want to pay for coverage that doesnt exist.

waveform06:
One thing I forgot to mention in the above. Is that as far as people who have looked at it can see, the group of wallets receiving the HNT income from the big list of asserted data-onlys are actually paying for the data transfer going thru their hotspots. It appears to be basically moving HNT from one set of wallets to another set via DC transfers for unknown reasons.

ilovespectra:
This was noticed by some members in the Discord and turns out it could be some form of money laundering, which is high effort - low result, but possible. It’s one of the only explanations since the DC is being paid for. There is a possibility a business is employing a large scale solution, writing the expense off as a business expense, and pocketing the HNT earnings to recoup personally money spent by their business. This is technically illegal but not within the jurisdiction of the network to try to interfere with. The tax man will be their enemy, not the network. As a permissionless network, I think it’s important to not interfere with network users, but simply make cheating more complicated. It sounds like an assertion requirement does not throw the stick in the spokes I’d hoped. 

I hear your previous concerns as well, about which hotspots are handling the data, and the absence of impact simply asserting may have. If there is any additional refining or suggestions that could be made that could bring this more into the scope of a required fix, I’m open to it. If my solution is half baked and will not positively impact the network, I won’t take any offense to this HIP being revised or discarded. Just wanted to help if I could. Thanks again for your feedback.

ilovespectra:
After discussion with the community, based on how exactly DO hotspots communicate with the network, I'm afraid this HIP doesn't hold any water. If someone wants to take the bones and make it something, be my guest, I'm happy to help however I can but I believe this location assertion HIP would not have the intended effect. Thanks for the time and the feedback, crew.

Issue #458, opened 2022-07-26
HIP 69: Re-assertion Fee Reduction


edakturk14:
# HIP 69: Re-assertion Fee Reduction

- Author(s): @therealjohnmac50
- Start Date: 2022-07-24
- Category: Economy 
- Original HIP PR: https://github.com/helium/HIP/pull/453
- Tracking Issue: [This](https://github.com/helium/HIP/issues/458) 
- Status: In Discussion 

# Summary

The average daily mining reward per hotspot is currently equal to around 50,000 DC; and it costs 1,000,000 DC to reassert a hotspot's location. 

As a result, it would take nearly a month for the average hotspot to earn enough to pay for one location reassertion.

Therefore, if we are trying to expand the network and want hotspots in saturated areas to move into less saturated areas (or moreover in lone wolf-like areas), the current reassertion price is not viable and needs to be reduced.

# Rendered View 

https://github.com/helium/HIP/tree/main/0069-reassertion-fee-reduction.md

BruggiR:
One could make the amount of the fee dependent on the ratio of the number of hotspots in the old hex to the number of hotspots in the new one

TheRealJohnMac50:
> One could make the amount of the fee dependent on the ratio of the number of hotspots in the old hex to the number of hotspots in the new one. two examples:
> 
> 1. old hex 10 hotspots , new hex 2 hotspots --> fee =1/5 of standard fee.
> 2. old hex 5 hotspots, new hex 5 hotspots --> fee = standard fee

This is one of many fantastic idea. Just like the rest, it is truly a brilliant idea. However, the technical requirements to make this possible is on an extremely grand level, as opposed to simply cutting the current reassertion fee in half. Therefore, for the sake of technical efficiency

abhay:
I agree that the technical implementation of this HIP is _just_ a chain variable change. Specifically it's updating the value of `staking_fee_txn_assert_location_v1` to `500000` (from current value: `1000000`). 

I'm concerned that halving the cost of assertion, however, with no other controls will only give arbitrageurs a cheaper way to adjust their asserted location (potentially more frequently) and increase the number of `assert_location_v2` transactions on chain which increases chain load. 

I'd be in favor of something aligned with what @BruggiR's recommendation. I think someone should do some more analysis but a framework like:

1. First N location asserts paid for by the manufacturers as they are today. (1 or 2 is the case in the wild)
[...]

PR #589, opened 2023-03-19
Iot assertion fees

charlesfayal:
Updating IoT Gateway assertion fees

### DIFF
--- a/0000-iot-assertion-fees.md
+++ b/0000-iot-assertion-fees.md
@@ -1,0 +1,71 @@
+- Start Date: 2023-5-19
+- HIP PR: <!-- leave this empty -->
+- Tracking Issue: <!-- leave this empty -->
+
+# Summary
+[summary]: #summary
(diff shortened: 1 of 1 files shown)


vincenzospaghetti:
Hi @charlesfayal - the community recently passed HIP 69 #458 to lower fees for 3 months, but it is not a permanent fee reduction, and the going consensus is that we **should not reduce fees permanently.** Is your proposal to reduce this permanently?

If so, you need to make the argument as to **WHY** this will benefit the network. Reducing fees has a direct correlation to a LOWER DAO Utility Score, which would not be a positive outcome for the IOT subDAO. 

Can you spend some more time on this HIP? Please reach out to me directly if you would like some help editing.

waveform06:
@hiptron please close this HIP 
agreed with author

abhay:
Closing on behalf of HIP editors.

PR #590, opened 2023-03-20
Adds Mobile Mapper Rewards HIP

meowshka:


vincenzospaghetti:
@meowshka this HIP is missing a few sections. Can you add:

# Drawbacks
Why should we not do this?
What problems could occur if we do this?

# Rationale and Alternatives
This is your chance to discuss your proposal in the context of the whole design space. This is probably the most important section!
Why is this design the best in the space of possible designs?
What other designs have been considered and what is the rationale for not choosing them?
What is the impact of not doing this?

# Unresolved Questions
What parts of the design do you expect to resolve through the HIP process before this gets merged?
What parts of the design do you expect to resolve through the implementation of this feature?
What related issues do you consider out of scope for this HIP that could be addressed in the future independently of the solution that comes out of this HIP?
Are there dependencies, milestones, or dates that need to be met for this HIP to succeed?

# Deployment Impact
Describe how this design will be deployed and any potential impact it may have on current users of this project.
How will current users be impacted?
How will existing documentation/knowledge base need to be supported? Any content to change at http://docs.helium.com/?
Is this backwards compatible? Can this HIP be undone?
If not, what is the procedure to migrate?

# Success Metrics
What metrics can be used to measure the success of this design? Are any new ETL reports needed to measure the success?
What should we measure to prove a performance increase?
What should we measure to prove an improvement in stability?
What should we measure to prove a reduction in complexity?
What should we measure to prove an acceptance of this by its users?

meowshka:
@vincenzospaghetti done with latest commit https://github.com/helium/HIP/pull/590/commits/c81e815d351bf087254d42194a2a51967360219e

PR #606, opened 2023-04-04
Minimum onboarding fee for subDAOs

mawdegroot:
rendered view: https://github.com/mawdegroot/helium-hip/blob/mg/minimum-onboarding-fee/00xx-minimum-onboarding-fee.md


mawdegroot (commented):
Thank you for the edits. I think we're ready to push this to the next step of the HIP process.

PR #637, opened 2023-05-10
Adds clarification to the grace period that allows unrewarded data

meowshka:


vincenzospaghetti:
added clarifying language

PR #647, opened 2023-05-12
0000-astley-incentive.md

demogutentag:
The People's Network has known each other for so long
Your Hotspot's been aching, but is too shy to beacon
Inside, `gateway-rs` knows what's been going on
We know the goal and we're gonna build it


demogutentag:
if its not clear, this is for a demo, [please dont merge this](https://www.youtube.com/watch?v=dQw4w9WgXcQ&themeRefresh=1)

PR #653, opened 2023-05-14
0000-Create-IoT-Hotspot-Hex-Limit.md

heatedlime:
This Helium Improvement Proposal (HIP) suggests adding a limit of how many IoT hotspots can be placed in one size res8 hex to two (2) Hotspots, and removing the transmit scale system created in HIP 17.

### DIFF
--- a/IoT Hotspot Hex Limit
+++ b/IoT Hotspot Hex Limit
@@ -9,15 +9,17 @@
 
 
 ## Summary:
-This Helium Improvement Proposal (HIP) suggests adding a limit of how many IoT hotspots in a single hex can earn Proof of Coverage (PoC) rewards to two (2) Hotspots, while the top Hotspot earning full rewards, and the second top Hotspot earning half rewards. Further, this HIP would rermove the transmit scale system established in HIP 17.
+This Helium Improvement Proposal (HIP) suggests adding a limit of how many IoT hotspots in a single hex can earn Proof of Coverage (PoC) rewards to two (2) Hotspots, while the top Hotspot earning full rewards, and the second top Hotspot earning half rewards. Further, this HIP would remove the transmit scale system established in HIP 17.
+
 
 ## Motivation:
 HIP 17 was passed to reduce IoT rewards earned by transmitting or witnessing Hotspots in close proximity to each other, and discourage multiple Hotspots being deployed within the same hex. However, as we know, this did little to discourage the overcrowding of hexes.
-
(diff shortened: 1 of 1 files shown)


vincenzospaghetti:
I'll close this PR since we have a new one.

IoT Hotspot Hex Limit
```diff
@@ -0,0 +1,99 @@
...
+|HotSpot 9 | 4          | 0            | 1                   | 320.25        | 04/06/2022           |  0           | N/A              | 0          |
+
+|Hex E     |# of Beacons|# of Witnesses|Data Packet Transfers|Assigned Points|Hotspot Assertion Date|Awarded Points| % of Total Points|IoT Rewarded|
+|----------|------------|--------------|---------------------|---------------|----------------------|--------------|------------------|------------|
+|HotSpot 10| 4          | 39           |  4                  | 1,491         | 01/26/2021           |  1,491       | 38.94%           | 38,944.76  |
+
+
+In some instances, deployers may not set up Hotspots optimally for IoT rewards, and only intend to use the Hotspot for MOBILE or IoT data transfer purposes. Therefore, this HIP also suggests the implementation to effectively opt out of IoT PoC activity (witnessing and beaconing) for all IoT hotspots. The cost to opt out or opt back in for PoC activity will cost 100,000 DC ($1). In instances where Hotspots are “opted out” of PoC activity, that Hotspot will not generate any points for witnessing and beaconing.
```
samgutentag, line 80: Is the intention here to opt out of PoC but still transfer data? or is it effectively a "turn off IOT completely" option?

heatedlime:
> I'll close this PR since we have a new one.

Can you re-open this? This is a new one specifically for IoT. The other one was for MOBILE.

vincenzospaghetti:
My mistake! I see you closed the other duplicate!

IoT Hotspot Hex Limit
heatedlime: I'll make it a little more clear once this is re-opened, but the idea is to just turn off witnessing and PoC, but still allow the transfer of data. You will NOT be able to opt out of the transfer of data, since that's pretty much what the network was built for. 

This idea comes from back when MOBILE radios first came out, people would set up their freedomfi gateway, but not assert it in the same hex, or re-assert it in the same hex because they don't want it to mess up their transmit score. And a lot of people just used the stock antenna that would come on it, and didn't plan to optimize the setup since they were just using it as a mobile gateway. 

This effectively allows someone to use a gateway just as a MOBILE gateway, so it doesn't mess up the Hotspot count in its asserted hex.

hiptron:
Hi @heatedlime can you make sure to add the same feedback on the mobile hex hip to this one?
#654

heatedlime:
Closing for now.

PR #678, opened 2023-05-30
Merge patches 3 with main?

KeithRettig:
I need some lessons.  I do not get what I am doing.


KeithRettig:
merging up to main I think

PR #680, opened 2023-05-30
Keith HIPs (3 in total)

KeithRettig:
I apologize about this.  I didn't realize that it was better to keep HIPs separated into different branches.  Here are all three of my HIPs.  They have been edited and formatted to my understanding of how Valerie wants it (as per some editing requests earlier).  I think this pull request will in effect create a second HIP for Service Provider Rewards.  The versions of all three HIPs in this pull request are the most up-to-date.


vincenzospaghetti:
This is totally fine! Looks good! Will add some comments on the other files for your review.

0000-lower-minimum-device-onboarding-fees
```diff
@@ -0,0 +1,113 @@
+# HIP XX: Lower Minimum Device Onboarding Fees
```
vincenzospaghetti, line 1: If you're going to resubmit this HIP (HIP 81), you need to address WHY you're resubmitting HIP 81 and what you changed from the previous time. 

Be sure to address that although HIP 81 did not pass, we are resubmitting the HIP with X changes and believe it is still necessary because of A, B, C problems that exist in the network.

vincenzospaghetti (changes requested):

0000-lower-minimum-device-onboarding-fees
KeithRettig: Adding why it is believed that the slight modifications to HIP81 should allow the HIP to pass.  I am of the position that the problems being addressed are already part of the content.  They are addressing exactly the same issue.

KeithRettig:
I hope I am doing this correctly.  I think you might want to do this one before you work on the pull request for the changes to the Double-the-value HIP submission.

KeithRettig:
Some quick edits requested by the group as a result of discussing with capcom.

KeithRettig:
@abhay ?
This pull request contains the three HIPs I wrote (two of which were asked of me to do).

PR #683, opened 2023-06-02
Update 0082-helium-mobile-service-provider.md

Siegfried-B:
Changing a potentially incorrect phrasing. The incorrect phrasing introduces a contradiction to HIP 79:

HIP 79 says SPs aren't able to redistribute Mobile rewards to subscribers (for legal reasons)

The original phrasing here in HIP 82 can be easily read as "HM pays subscribers Mobile tokens that HM temporarily owned" - thus contradicting the statement in HIP 79 that SPs can't do that.

But what actually happens (in my understanding is): Having a subscription is a prerequisite to do discovery mapping and to get MOBILE rewards out of the mapper rewards bucket. Thus HM does not pay rewards, but being a subscriber of HM only enables discovery mapping and getting rewards for discovery mapping. 

The suggested change to the HIP reflects this.

### DIFF
--- a/0082-helium-mobile-service-provider.md
+++ b/0082-helium-mobile-service-provider.md
@@ -21,7 +21,7 @@ As a Service Provider member of the Helium Mobile subDAO, Helium Mobile will act
 
 This HIP affects:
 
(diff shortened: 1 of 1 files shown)


vincenzospaghetti:
Looks good!

PR #686, opened 2023-06-05
Hip82 keith 20230604

KeithRettig:
Changes made to opt-out v opt-in, deleted the cutoff rate calculation paragraph, and added in a requirement for a yearly subscriber usage report.  This should be done after Boris's changes are implemented.


KeithRettig:
resolved the conflict as you asked.

PR #700, opened 2023-06-12
Create 0000-lower-minimum-device-onboarding-fees

KeithRettig:
Create new HIP pull request under request of @abhay.  Supersedes previous pull request.


waveform06:
(1)"HIPs 51-53 have defined the onboarding fees as $40 per hotspot" - actually only HIP53 defined an onboarding fee (and an assert fee). HIP52 inherited the fees from existing hotspots. HIP51 talked about activation fees. Does this HIP confirm then that MOBILE hotspots require a $40 Onboard and $10 Location assert fee upon this HIP implementation?
(2) On the initial mention of the VDA factors and DAO Utility score it's probably a good idea to state this refers to HIP51 section Omni Protocol PoC Incentive Model or refer to your Rationale and Alternatives section.
(3) Grace period section needs further clarification.
What is happening with the location assert fee for the existing hotspots, ignored?
Is it FreedomFi and Bobcat 5G specifically or are these the only 2 makers assumed will have shipped 5G units by September 4th?
I don't really understand what is happening in the 2nd paragraph under the grace period section. Are you saying that unless paid by someone, the MOBILE treasury is going to back pay the DC onboard fees from HNT being earned from the extra HNT entering the treasury that it receives from pretending the onboard fees had already been paid? Why would anyone pay MOBILE onboarding fees if this HIP says it will pay them from the treasury if they are not paid?
(4)Onboarding devices section. Last sentence. Actually, Data Only hotspots do count towards the DAO Utility score as if they had paid $40 not the $10 they did pay. Implementation of this HIP and calculating the actual onboard fee will fix this.
(5) Active devices. You appear to be defining a hotspot passing data as "inactive" and this could be very confusing for network customers. A better way would be to device Active devices as POC Active devices.
(6) Multiple subDAOs in a single device - are assuming that 5G hotspots asserted for IOT but not Location Asserted (Earning MOBILE) are dual function hotspots
(7) Drawbacks. The DAO Utility score code will need to be completely new code, as it moves from a summed number X a variable to a database of previously paid values that are looked up each epoch and then summed based on PoC activity. Who has confirmed they will write this code? Or who are you asking to write this code?

KeithRettig:
> (7) Drawbacks. The DAO Utility score code will need to be completely new code, as it moves from a summed number X a variable to a database of previously paid values that are looked up each epoch and then summed based on PoC activity. Who has confirmed they will write this code? Or who are you asking to write this code?

This is not how it has been described before.  The summed number is what?  The number of active devices within that epoch?  Or a total of the number of hotspots that just keeps getting added to each epoch?  If it is the total amount of onboarding fees ever paid, then I am in.

KeithRettig:
> (1)
> (2)
> (3)
> (4)
> (5)
> (6)
> (7)

Did you comment on HIP81 and they ignored you?  I thought I was going to get away with few edits given I basically just regurgitated HIP81.  :)

waveform06:
I had chats with Noah about HIP 78/80/81
The DAO Utility score is just the variable for DAO onboarding multiplied by the number of "active" hotspots in that epoch.
Its really simple, do a count and then multiply by a constant variable.($40) 
So currently Data Only IOT hotspots (if "active") get counted as $40 not $10 as it doesn't know there is a difference as Noah didn't know there was a difference in onboard fees for a Data Only.
If we start recording the actual amount paid, then we need a database of what value onboarding fee was paid by each hotspot and the DOA utility score has to lookup the fee paid by the hotspot for each active hotspot and then sum them. It was too late for this to be implemented before the migration day (needed a new audit) which was why I was always pushing for either a $0 MOBILE onboarding fee or someone to pay the back fees for the migration.

gateholder:
Thats kinda why I think this is a good long term solution but not a good short term @KeithRettig can I propose that we combine the short term fix from PR 697 by removing the Active devices*40 and replacing with just active devices (being gateways as this is already done)

KeithRettig:
> Thats kinda why I think this is a good long term solution but not a good short term @KeithRettig can I propose that we combine the short term fix from PR 697 by removing the Active devices*40 and replacing with just active devices (being gateways as this is already done)

Isn't the problem with this approach that a subDAO can game it by making their definition of "active device" be something inexpensive and to also define their onboarding fees as very low?  This would allow such a subDAO to get an out-sized amount of the HNT emissions each epoch.

gateholder:
> 

What I am proposing is that this is temporary, basically to solve the problem Immediately just chop the 40 out and then when your coding is done then it gets superseded?

abhay:
What's our plan with this PR? Close?

KeithRettig:
close it.  I will edit if it becomes useful and create a new pull request

abhay:
Closing as requested.

PR #712, opened 2023-06-19
Update 0088-adjustment-of-dao-utility-a-score.md

heatedlime:
Changed the whole premise of HIP 88 to change the A factor to be calculated by taking the # of active devices multiplied by the onboarding fees those active devices paid.


hiptron:
Merging author's submission.

PR #716, opened 2023-06-20
Driving value into the iot network through iot token payments

heatedlime:
This HIP proposes changing the method of payment for Location Assert Fees on the IOT subDAO from Data Credits (DC) burned from HNT to IOT tokens. The primary purpose of this change is to drive value into IOT tokens and provide them with a small amount of utility. This HIP does not propose altering any other aspect of Location Assert Fees.


heatedlime:
Closing for now to revisit idea later.

PR #718, opened 2023-06-21
HIP 89 Updates

heatedlime:
Significant edits were made to the original HIP 89.

PR #720, opened 2023-06-21
Hip 89

KeithRettig:
Made some edits based on Chris Ferebee's suggestions.


abhay:
Does this conflict in any way with #718?

ferebee:
It should be based off the content of #718, assuming that contains what Andy linked earlier in the channel.

ferebee:
Confirm, just did a local comparison.

KeithRettig:
Changed "funding" to "paying" (and other similar changes to try to clarify that one does not store money in the onboarding server.

waveform06:
My thoughts on this updated version are.
(1) Stakeholders include everyone affected. So that includes MOBILE token holders (value in HNT will start to increase),  IOT Token holders (value in HNT will start to decrease) 
(2) If the HIP is urging compliance with HIP-53 then it should also require the Location Assert fee to be set to $10 or say a update to HIP 53 is required that the Mobile Network no longer has an on chain Location Assert fee, seeing as its a "Given that there is no Location Assert operation in the MOBILE subDAO" (If the latter then Maker Apps need updating)
(3) Ref "what to do with the excess funding (10 USD per Hotspot) presumed to be paid into the Onboarding Server for Location Assert is outside of the scope of this HIP." As mentioned before, there is no funding into an Onboard server and all funds are traceable in maker wallets so no presumptions about funds should be made.

abhay:
I think this branch needs a rebase since it appears that it was based off an older branch. the filename change doesn't seem to be working correctly.

KeithRettig:
Feel free to close this pull request as the requested action above has been completed.
Going to try to close it myself...

PR #728, opened 2023-06-28
0000 -excess-mobile-allocation-budget

heatedlime:
This HIP stipulates what the 15 Billion of the 20 billion MOBILE currently under the Helium Foundation custody will be used for.


waveform06:
There are 4 updates on this PR
Looking at 0000-MOBILE-Network-Onboarding-Fee, this is too similar named to HIP 89: MOBILE Network Onboarding Fees by the same authors. Can you rename more indefinably please
PR does not confirm if the existing Hotspot Onboarding fees are being replaced or added to. If they are being replaced then Stakeholders should include the Existing Hotspot vendors who will no longer pay fees if HIP 89 is approved.
PR does not defines who actually pays the new fees at the point of onboarding, 
PR does not come with code or a commitment from any organization to write the code,  just a lot of  "Need"s that the Foundation creates code. The Foundation cannot create all the code needed on the Output devices, the Hotspots, nor the code needed on the Maker Apps so cant be held to a deadline that requires these. Deadline needs to be estimated by the all the stakeholders who will implement. 
Unresolved Questions: How does this integrate, react, affect HIP 89 and HIP 88 and how does this affect the success metric?

waveform06:
Looking at 0000-MOBILE-Oracle-Hex-Boosting.md
Is this a complete replacement for the earlier PR?
PR does not come with code or a commitment from any organization to write the code in the deployment section.

waveform06:
Looking at Change IoT Hotspot Reassertion Fee

Stakeholder to add: IOT Hotspot Makers holding DC reserves that will not be consumed by the expected Hotspot deployment before this HIP is implemented. 

If the success metric is that burning IOT reduces IOT supply (increasing IOT value against HNT, then the Motivation statement could be described as incorrect in that the burning of HNT which reduces HNT supply, increases HNT value which increases IOT value.

PR #738, opened 2023-07-09
HIP 91

maxgold91:
[Rendered](https://github.com/maxgold91/HIP/blob/IOT-Assert/00xx-data-driven-extension-reduced-iot-assertion-cost.md)


abhay:
Can you remove the other file from this PR?

PR #749, opened 2023-07-24
HIP 94: Response Time Windows for Witness Rewarding

disk91:
…o 14 randomly selected hotspots received in a given time windows


waveform06:
Please change 92 to XX. HIP authors do not allocate numbers to HIPs.
See the HIP template https://github.com/helium/HIP/blob/main/0000-template.md

The HIP has no sections for:
Rationale and Alternatives
Unresolved Questions
Deployment Impact
(Who will be writing the code for this?)
Success Metrics

waveform06:
Describing simple scenario numbers will help the understanding as well.
Eg in a scenario with 50 witness receipts of which 10 fail to meet good LoRaWAN-grade timing constraints.
Pre HIP 83, it was a random selection of 14 of all the 50
Post HIP 83, it selected the fastest 14
Post this HIP, it is a random selection of the 40 out of all the 50, that meet minimum good timing constraints.

disk91:
Scenario still missing but template format should be better and multiple illustration have been added

XX-Response-Time-Windows-for-Witness-Rewarding.md
```diff
@@ -0,0 +1,188 @@
...
+Hotspot with the worst locations, indoor, in the city center, gets the best response time experience 
+with direct fiber connectivity.
+
+We have seen that depending on the hardware of the hotspot, the Witness processing time is largely dependent 
+on the packet signature as described in the [appendix](#ecc-signature-impact). As the signature is delegated to an hardware 
+ECC chip for most of the deployed hotspot, the processing time depends on the hardware solution soldered. 
+Even if firmware can be improved, the current solution disqualifies certain hardware whatever is the hotspot owner's efforts. 
+
+More generally speaking, it disqualifies lower-end hardware like light-hotspots based on micro-controllers as, by definition, their capacity to process the same thing than an hotspot based on CPU is lower. This is nonsense as these hardware have a 
+better fit for stability, long-life, energy saving, lower cost, and should be privileged long term.
```
mawdegroot, line 51: Is there any proof backing this up? Because current light hotspots in the field do not exhibit the behavior you describe.

XX-Response-Time-Windows-for-Witness-Rewarding.md
```diff
@@ -0,0 +1,188 @@
...
+This illustration is based on a real exemple, but is a general illustration for hotspot placed on cell-towers and high elevation point. The hotspot used in the illustration as model is Attractive-Olive-Cybord, located in Clermont-Ferrand, France. This hotspot is deployed on high elevation point. Coverage can be seen on mappers.helium.com, with 80km coverage from North to South. 
+
+## Suburb Valuable Coverage
+
+The hotspots located in suburb and above, in blue in the illustration, are expending the network coverage with uniq 
+coverage zone and are in competition with the hotspot inside the city. They are also the hotspot that are opening the PoC rewarding to the next hotspot out of the city and suburb. 
+
+![Suburb illustration](XX-response-time-windows-for-witness-rewarding/suburb-coverage.png)
+
+These hotspots does not get benefit of the fastest Internet connection as their fiber connectivity will pass through the city center to reach the main Internet highways. For most of them the fiber connectivity will not be available and they are going to rely on xDSL connectivity before reaching the ISPs fiber Internet backhall.
```
mawdegroot, line 112: That's a pretty big assumption without supporting evidence. I don't think this holds true in general, but it might hold true in the particular example you show.

XX-Response-Time-Windows-for-Witness-Rewarding.md
```diff
@@ -0,0 +1,188 @@
...
+## Packet Processing and LoRaWan time constraints
+
+The Helium Packet Router (HPR) is accepeting all the coming packets up to the limit of the max_copies set on the route or eui in 
+the config service. First come, first paid. Only for roaming the HPR is applying a time limit. The time limit for a packet is 
+decided by the LNS after the HPR. This time limit for the LNS is a LNS setup and can vary for each of the LNS.
+
+LoRaWAN time constraints are the following:
+- UPLINK first copy arrival has no time constraint, next copies are withing the defined LNS deduplication time windows.
+- JOIN REQUEST needs to be responded within 5s for RX1 (same frequency, standard power) or 6s for RX2 (other frequency, potentially higher power). LNS decides of the selection between RX1 and RX2 dynamically, according to the time available.
+- DOWNLINK REQUEST / ACK needs to be responded within 5s for RX1 (same frequency, standard power) or 6s for RX2 (other frequency, potentially higher power). LNS decides of the selection between RX1 and RX2 dynamically, according to the time available.
```
mawdegroot, line 125: Downlinks are 1 second RX1 and 2 second RX2.

XX-Response-Time-Windows-for-Witness-Rewarding.md
```diff
@@ -0,0 +1,188 @@
...
+
+There is no reasons to prefer RX1 vs RX2, most of the implementations try to reach RX1 first, but RX2 provides different advantages, in particular in Europe where the duty cycle on RX2 is better and the higher power makes more chance to reach the device.
```
mawdegroot, line 127: If RX1 suffices it certainly has benefits over RX2 because it uses the same channel (= more bandwidth available) and the device can return to sleep faster (= energy saving)

mawdegroot (commented):
I would like to see more data beyond anecdotal evidence to back up some of the claims made in this HIP. 

It's also important to note that first to respond is exactly how data packets are currently bought by the network and this proposal removes the alignment between proof of coverage and data transfer. The witness selection criteria only play a role in areas where the density is already 10 hotspots higher than the network's target (target = 4, max witnesses = 14).

With the average time between the fastest and the 14th witness being around 250ms this proposal seems primarily focussed at rewarding 4G hotspots on a mountain above all else. For example the Attractive Olive Cyborg hotspot is currently being rewarded roughly 3x the network average while changing to the randomized way would bring it back to 6-7x the network average. If we compare this hotspot to the hotspots in the so called city center on "fiber connections" which are receiving roughly 0.5x the network average it is hard to argue that the hotspot with "highly valuable coverage" isn't rewarded appropriately.

Finally, it would be interesting to see how this would affect current hotspots in a way similar to how HIP83 visualized it on a per hotspot basis.

XX-Response-Time-Windows-for-Witness-Rewarding.md
disk91: Evidence does not really need to be shown.MCU computation power is lower than CPU  computation power. Code on MCU may be better optimized someway ... but gateway-rs code is the same for all. Currently time is mostly spent on ECC signature witch is equivalent and hide the other difference. Let's be a bit scientific or prove me that there will never be a difference of computation power in favor of CPI based system compared to MCU based system.

XX-Response-Time-Windows-for-Witness-Rewarding.md
disk91: This is also an evidence, that the way the networks are build.

XX-Response-Time-Windows-for-Witness-Rewarding.md
disk91: it also have larger negative impacts : it has a higher risk of collision because it create a larger use of the same bandwidth (so basically you have less bandwidth) , is has a 1% duty cycle, really impacting the gateway duty cycle shared across all the devices and the limited power in Europe is a great disadvantage with a larger reception loss due to the unbalance noise level and antenna radio gain at the device level. This conduct to retransmission with a higher cost than the wait for 1s. Downlink usage in a correct implementation is limited to a couple per day and at the end the global power impact is limited. This is what an experienced device developper will tell you about it. RX1 / RX2 is a LNS choice not a network choice. As the HIP demonstrate it, the factor impacting the choice between RX1 and RX2 in a windows of 200ms is at first related to the distance between the device and the LNS.

XX-Response-Time-Windows-for-Witness-Rewarding.md
mawdegroot: An evidence of what? If you're saying that fiber will only pass city centers and nowhere else then I don't think your statement holds true.

XX-Response-Time-Windows-for-Witness-Rewarding.md
disk91: Concerning the global remark, PoC reward coverage, if we take that exemple - and this is just and exemple as you requested exemples, to illustrate a general principle that apply to any cell-tower hotspot - this hotspot is 60km radius (80 for real) covering 60*60*3.14 km2 of territory. This is to be compared with regular good hotspot 10*10*3.14 km2. basically the coverage is 36x more. If we compare to the average coverage witch is more about 1km range 1x1x3.14 km2 is about 3600x. You can consider that x7 in reward well paid and I won't agree in this.
The question is not about the amount of reward this particular hotspot have. The question is why this hotspot get a penalty that can conduct some of the them to get 0 reward, compared to the one in city center, with fiber, providing a small coverage than get a benefit of the current HIP-83 situation.
HIP83 did not show any evidence of it's positive expected impact when written. So please show us evidence that the rewarding is, since HIP-83, better balanced to extend the network coverage. That way we will have something interesting to discuss.

XX-Response-Time-Windows-for-Witness-Rewarding.md
HeliosHyperion1111: > An evidence of what? If you're saying that fiber will only pass city centers and nowhere else then I don't think your statement holds true.

https://fttx.gr/

that's a basic example to your ''asking for evidence'' claims,
if you see the map, you can see cabinets of fiber, the center of city has the most dense fiber infrastructure while rural are barely covering anything, the unique coverage is where there is no fiber.
hope that helps to draw a picture what your HIP 83 actually did.

HeliosHyperion1111:
> I would like to see more data beyond anecdotal evidence to back up some of the claims made in this HIP.
> 
> It's also important to note that first to respond is exactly how data packets are currently bought by the network and this proposal removes the alignment between proof of coverage and data transfer. The witness selection criteria only play a role in areas where the density is already 10 hotspots higher than the network's target (target = 4, max witnesses = 14).
> 
> With the average time between the fastest and the 14th witness being around 250ms this proposal seems primarily focussed at rewarding 4G hotspots on a mountain above all else. For example the Attractive Olive Cyborg hotspot is currently being rewarded roughly 3x the network average while changing to the randomized way would bring it back to 6-7x the network average. If we compare this hotspot to the hotspots in the so called city center on "fiber connections" which are receiving roughly 0.5x the network average it is hard to argue that the hotspot with "highly valuable coverage" isn't rewarded appropriately.
> 
> Finally, it would be interesting to see how this would affect current hotspots in a way similar to how HIP83 visualized it on a per hotspot basis.

that's very anecdotal claims, the same way you pointed at his scenario ''is cherry picked'' the same way you cherry picked 1 hotspot that makes x3, while if i bring you 3 4 hotspots that are in the center of city in a HEX of 5 hotspots( 1 of them claims 3k IOT the rest fall into the category of 2k and the lowest is 1k) 
that's plenty of info to gather and see that the HIP 83 did more damage than good.
while looking at it on suburban/urban areas, if you call x3 of the network as 600 IOT per day, i can just laugh because this is hypocrite of you to only reward 600 IOT a hotspot that provides UNIQUE coverage while giving everything to the useless coverage center city hotspots, who made you the earnings police by the way?

PR #761, opened 2023-08-16
Adds details and clarifications to HIP93

meowshka:


abhay:
merging on behalf of the author but the HF bits in here probably needs to be discussed :)

PR #763, opened 2023-08-23
Adds paragraph describing high level Hotspot Vendors review process

meowshka:


abhay:
merging on behalf of authors.

PR #769, opened 2023-08-28
Clarifies AP limit per hex effect on data transfer reward eligibility

meowshka:


abhay:
Merging on behalf of authors.

PR #771, opened 2023-08-28
Mentions current PoC algo that will be used before launch of HIP74

meowshka:


abhay:
Merging on behalf of authors

PR #773, opened 2023-08-30
Create 0000-Compensation-for-Network-Disruption-on-2022-11-16

jaym2518:
This HIP proposes a compensation mechanism for Hotspot hosts impacted by the network disruption that occurred on 2022-11-16.

### DIFF
--- a/0000-Compensation-for-Network-Disruption-on-2022-11-16
+++ b/0000-Compensation-for-Network-Disruption-on-2022-11-16
@@ -1,0 +1,45 @@
+# HIP xx: Compensation for Network Disruption on 2022-11-16
+Author(s): @jaym2518
+Start Date: 2023-08-30
+Category: Economic
+Original HIP PR:
+Tracking Issue: 
+
+## Summary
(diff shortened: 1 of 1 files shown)


waveform06:
Jay, there have been other HIP based requests for compensation for network disruption in the past and non have been approved to go further. 
One reason is that if one is allowed then it will surely be followed with multiple HIPs even for compensations for the most minor disruptions and the resultant action will be that the devs are deployed purely to investigate and calculate the missing tokens and then to create the code and manual blockchain work to cover these multiple one-off distributions and thus they will spend less time on the tasks that prevent any further disruptions
Another reason is that you have left it in the hands of the Foundation to determine the source of the funds. The Foundation will not fund this compensation from its own funds, and the community agrees that one one entity cannot be forced by another (foundation) to pay funds to another or this again will lead to frivolous HIPs. This leaves the only other alternative for the tokens to be to mint new tokens. 
And minting new tokens will need its own HIP and dev support (as your compensation HIP if passing would be the first of many) - which it will not get. As this has been refused in the past for compensation requests, as it would involve changing the max emissions and notifying exchanges and financial institutions saying we are changing emissions for minor issues. It would also require more dev time to implement and additional costs for Solana smart contract audits.
As your drawback says "This could set a precedent for future compensations which may be exploited." and will come at increasing impact in time and costs for probably a decreasing return in compensations.
To sum up Helium Inc and the Helium Foundation have effectively declined to identify a source of funds or provide Helium Core Dev time in the past for similar proposals. 

This is very different to what HIP 92 proposes, which was a missed payment of a specified amount for a specific action to a specific group, it is not a compensation for an unknown action. And again no wallet is being forced to pay another and no new emissions are created. A unique solution for a unique situation.

PR #783, opened 2023-09-19
HIP 92: Move datasets into HIP repository

abhay:
Since we don't want to depend on external links that can be updated, I've moved the linked documents into the HIP repository for now. In the future we could also use other immutable or easy to trace storage like ipfs links. 

cc: @capjbadger007 @leogaggl

Let me know if this is an acceptable change.


leogaggl:
Making this part of the actual repo rather than external Google Drive is a good idea.

Thanks - all good with that change.

PR #799, opened 2023-10-25
making it clear that only valid heartbeats are counted

riobah:


hiptron:
Closed in favor of #800 as requested by the author.

PR #805, opened 2023-11-14
s/latency/random/g

BenoitDuffez:
Rendered view: https://github.com/BenoitDuffez/HIP/blob/s/latency/random/g/xxxx-s-latency-random.md


xxxx-s-latency-random.md
```diff
@@ -0,0 +1,61 @@
...
+
+As a reminder of said value from the first paragraph:
+
+* routing packets: unless they are unreasonably slow (see table above), all hotspots provide the same amount and value of work (e.g. routing the packet from A to B)
+* witnessing coverage: the uptime, coverage, RF performance, reliability, power usage, data consumption, etc.
+
+It is very hard to assert a reliable KPI to assess this value.  
+Today, latency fails to do so.
+
+Short of a better KPI, randomness seems the fairest option.
```
madninja, line 52: Would one option be to actually use the above "at least this fast" as a minimum (i.e. KPI!) and then randomly reward within that set? vs full random 14?

BenoitDuffez: I have no opinion on 14 per se, or 14 vs everybody. Thus, I'd like this to be out of the discussion (e.g. keep the current 14).

Indeed, "at least this fast" is a good question. As highlighted in the table, the "at least" is 2000ms if there is a class A downlink scheduled, and up to the application layer if there is not (not less than a few seconds even for the most dramatic use case).

Thus, depending on the downlink queue, the threshold could be 2000ms or say 5000ms if empty queue.

However, the backhaul latency isn't the only metric end-to-end, so headroom must be added and processing time given to the different Helium cloud nodes.

If I were to pick the threshold, I'd say 800ms RTT if there is a downlink, and 2500ms if there is none.

BenoitDuffez: Strictly speaking, there is:

1. uplink latency
2. cloud processing
3. downlink latency

At step 2 we know what the downlink time budget left is (RX window chosen, processing done). Thus we can at this step (with an additional processing headroom) select which hotspots actually could be chosen for a downlink (with an additional headroom to accommodate the downlink latency value and variance).

Hotspots could be eliminated before step 3 by assuming that the latency is symmetric.

(only if there is a scheduled downlink, otherwise the threshold could be relaxed to the 2500ms I mentioned above)

madninja: right.. (and yes I wasn't picky about the number of hotspots to be rewarded that fall inside the "fast enough" line). 

Would be good to work what you said into the proposal, since it actually adds some initial detail around the max latency value (or whatever the right name is)

BenoitDuffez: will do

BenoitDuffez: done, is that better?

xxxx-s-latency-random.md
```diff
@@ -51,11 +51,34 @@ Today, latency fails to do so.
...
+| Uplink with a scheduled class A downlink, target RX1  | 800ms     | Gives 200ms headroom for cloud processing (LNS handling, window selection, latency computation, etc)      |
+| Uplink with a scheduled class A downlink, target RX2  | 1800ms    | Gives 200ms headroom for cloud processing (LNS handling, window selection, latency computation, etc)      |
+
 ## Behavior
 
 For PoCs or data packets:
 
-1. wait an aggregation period
-2. collect hotspots that brought value (routing/witnessing)
+1. first packet received from first hotspot: check if there is a class A downlink pending, decide RX1/RX2, and compute threshold accordingly
```
madninja, line 81: I don't think this is realistic, unless I'm misunderstanding things.. PoC is beacons/witnesses which don't have downlinks so threshold can't be computed at this stage.

BenoitDuffez: Indeed. As said on discord, groot pointed out that we could set the bar for PoCs as to "who would have been able to do a RX1 downlink on that one", and thus merging this case into the "RX1 downlink" case.

madninja: This first point still needs to get removed.. the threshold can't really be dynamic in the current system

xxxx-s-latency-random.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Drawbacks
+
+* removing the race to latency and switching to lottery for those under a threshold, if threshold is too high, could lead to increased earnings for packet stuffers
+
+## Rationale and Alternatives
+
+Removing the race to latency is fair for hotspot owners because latency wasn't a HIP19 requirement when makers started selling hotspots.
+
+A *good* threshold is crucial to avoid packet stuffing. It needs also to be fair, e.g. not exclude any maker or future hardware (so-called *light hotspot*).
```
madninja, line 115: A good threshold is also important for ensuring low latency for data packets

waveform06:
@BenoitDuffez If you are satisfied with the goals of HIP-94 can we close this.
HIP-94 is awaiting confirmation if Disk91 can get someone willing to code it before it goes to vote.

BenoitDuffez:
If 94 gets to a vote yes we can close this.

hiptron:
Closing as stale. Please reopen when ready.

PR #817, opened 2023-11-27
0000-Reduce-IoT-Onboarding-Fees

mrfizzy99:
# Reduce-IoT-Onboarding-Fees

- Author(s): @mrfizzy99
- Start Date: 2023-11-25
- Category: Economic and Technical
- Original HIP PR: 
- Tracking Issue: 

# Summary
[summary]: #summary

  - This HIP aims to reduce onboarding fees from the IoT subdao by changing it from $45 (4,500,000 DC) to $10 (1,000,000 DC).
  - The First Location assert is still expected to be paid by the maker account, which is currently $5 (500,000 DC).
    - With the current set location assert of $5 (500,000 DC), this HIP will allow 2.5x the number of hotspots to be newly onboarded and asserted from maker accounts when compared to the current set fee values.
  - This HIP also aims to require makers to maintain dynamic and useful balance in their maker account for remaining first location fee. 

# Motivation
[motivation]: #motivation

For almost a year, the Helium Network has been plagued with Maker accounts running out of DC or SOL and not replenishing to an appropriate amount to fulfill their obligations under the MCC Maker Ethics Document. This has caused confusion and has been hurting the out of box experience that new users of this eco-system go through. The workaround solution that has been offered to the community has been that the Hotspot Operator would need to go through additional steps to burn HNT for DC into the Makers account, the equivalent of $50 (5,000,000 DC). Essentially repaying the onboarding fee and the location assert fee that has always been marketed and sold as 'included with' in the past from every maker in the ecosystem.

# Stakeholders
[stakeholders]: #stakeholders

Entire IoT network and hotspot operators and manufacturers, current and future.
  - We ask that thoughts about this HIP be directed to the Helium Community Discord Server.

# Detailed Explanation
[detailed-explanation]: #detailed-explanation

Onboarding fee gets set to $10 (1,000,000 DC), ZERO cost to get a hotspot onto the blockchain from the traditional $45 (4,500,000 DC) 
  - At current, this is a major blocker to any further deployment of IoT hotspots within the Helium ecosystem.  And for months now users have been told to just pay it themselves or wait for maker to refill their DC account.  The pay-it-your-way onboarding has gone on for to long and needs to be resolved.  

First assert fee remains expected and required from the hotspot maker.  
  - To keep DC useful and part of burn economics rolling in this equation, makers will still be required to pay for the first assert fee.  Currently set to $5 (500,000 DC). 

Requiring a minimum DC and SOL Balance: 
  - $100 (10,000,000 DC) and 0.2 SOL.  A bare minimum, enough for 20 first asserts at the current set location assert value, assuming no attempted or successful onboards within the past 7 days.
In the case that a maker account has dipped below the threshold amount $100 (10,000,000 DC) OR 0.2 SOL for a period of 7 days (168 hours), the maker will be warned warned within 24 hours via email communications and public Discord communications announcement by either the Foundation or a member of the MCC that they are under the threshold as determined by this HIP, and to replenish either DC and/or SOL within the next 7 days. If 14 days (336 hours) have passed since dipping below the threshold, their maker key will be temporarily suspended, and the maker will be given notice by the Foundation or MCC member that they have 14 more days to either replenish or the maker key will be permanently suspended. 28 days (672 hours) after dipping below this threshold, their key is permanently suspended. All maker accounts are subject to this minimum balance, including test and internal maker accounts. 

In the case that a maker runs out of 20 onboards in day 1, or less than 7 days.  Makers will also be required to keep a weeks average worth of DC and SOL to account for the onboarding and first assert fees on a weekly basis. They are encouraged to put forth more of course to prevent weekly discord pings and communications that they are under their set threshold. If the maker runs out of DC or SOL, in which onboarding and first asserts are unable to proceed, an weekly average will continue to be calculated by the number of unique hotspots that attempt to onboard during that rolling week. 

If a maker reaches a point at which no onboards or first asserts are possible, the maker will be warned within 24 hours via email communications and public Discord communications announcement by either the Foundation or a member of the MCC that they are under the threshold as determined by this HIP.  In both, the calculated threshold of DC and SOL required to meet their personalized minimum threshold based on the rolling 7 days will be mentioned.  In this special case, they are only given 7 days to fulfil the minimum threshold before the temporary suspension, and 14 more days till permanent suspension.

Examples of makers that have not had an onboarding problem, so we have clear data (11-18-23 to 11-25-23):
  - Sensecaps at the time of writing, has had 46 onboarded hotspots within the past 7 days.  Assuming this is a normal trend, the community can expect that maker to have enough for the next week. The maker Sensecap will be asked to have at least $330 (3,300,000 DC) and 0.66 SOL within their maker account as their minimum (46 weekly x $5 onboard) + ($100 min). 
  - Milesight at the time of writing, has had 5 onboarded hotspots within the past 7 days. Assuming this is a normal trend, the community can expect that maker to have enough for the next week. The maker Milesight will be asked to have at least $125 (1,250,000 DC) and 0.25 SOL within their maker account as their minimum.
  - We assume the $100 (10,000,000 DC) AND 0.2 SOL requirement is a bottom threshold for less active maker accounts with 0 onboards weekly as a bare minimum.

To Summarize:
  - Maker account threshold requirement:
    - If 0 attempted or successful onboards within a rolling 7 days: => $100 (10,000,000 DC) AND 0.2 SOL.
    - If >=1 attempted or successful onboards within a rolling 7 days: => Y (attempt/success onboards) x  Z (first assert fee / with sol txn cost)  + $100 (10,000,000 DC) OR 0.2 SOL.

  - Events x days after dipping below thresholds:
    - 0 Days - Warning of low DC or SOL if maker account is unable to onboard or first assert a single hotspot.
    - 7 Days - Warning of low DC or SOL to meet the required minimums.
    - 14 Days - Temporarily suspended maker key, regardless of remaining DC or SOL.
    - 28 Days - Permanently suspended maker key, regardless of remaining DC or SOL.

Ideally, popular makers will top up for the monthly expected onboards if they have not already. If makers are consistently warned that they are under thresholds within email communication and public discord space, it should be a signal to the community that the maker may no longer be interested in supporting their product within the Helium ecosystem in their current maker capacity. And it should also signal to the community that there could be a 7 to 14 day timer for last time they could be able to first assert and onboard the hotspot from that potentially exiting maker.

Once the maker is suspended, the community, along with the Foundation and MCC should communicate openly on the situation and start updating docs and material to reflect the maker walking away from the Helium ecosystem.  Because the onboarding fee burden has been set to $10 (1,000,000 DC) the hotspot can be openly used and the keys rolled for any other manufacturer for a lower fee, or perhaps a graveyard type maker account could be created for such hotspots without a supporting maker, and this could be managed or maintained by the Foundation and/or MCC in which users only have to pay for the first location assert after being rerolled to the graveyard maker account. Firmware support could be community-based or maintained by another maker (such as with Nebra balena fleet). 

# Drawbacks
[drawbacks]: #drawbacks

  - Reduces the burn effect on the ecosystem.

  - This could force some makers to completely ghost or disconnect from the Helium ecosystem, depending on their situation, and could cause a number of maker keys to be suspended 28 days after this HIP is installed.

# Rationale and Alternatives
[alternatives]: #rationale-and-alternatives

This proposed HIP is the best solution for the IoT ecosystem going forward into the future to allow it to remain easy and friendly to use. It will visually reduce the DC and SOL cost for makers to have in their maker accounts, and will be easier for them to maintain an appropriate balances for the onboarding demand within a rolling time frame.  

  - Alternative # 0:  Set the first assert to always be paid by the Hotspot deployer, while also setting $0 (0 DC) onboarding fee.
    - This will inadvertently make DC in maker accounts have no purpose; see #Unresolved Questions .

  - Alternative # 1: HIP 95: Self-Onboard Hotspots After Maker Exit.
We authors do not see this alternative as a viable long-term fix or even beneficial to the network in any way. It does not stop a maker from "Ghosting" the network, with only an acknowledgment every so often saying that they are still alive without any commitment to pick between officially exiting or refilling DC/SOL balances for the initial onboarding process.  It also still holds current makers bound to the Ethics document while allowing other makers to get a parachute out of the network with less public backlash while still pocketing the fees that were promised to be included with the hardware they sold (In the way that one could say that if HIP-95 was approved, it is a fair, just, and respectful way for a maker to exit without consequences of what fees are due to the network by the exiting makers). 

  - Alternative # 2: Completely requiring onboarding fees and first assert from the hotspot operator, an additional $50 (500,000 DC) cost to the operator after opening the box of a new hotspot.
All onboarding fees and assert fees are paid by the user, no matter the maker.  $50 (5,000,000 DC) will be required from the operator of a new hotspot from a approved maker with an active maker key. We want to assume that no one wants this option to repay what was already supposed to be paid. But we list it here as an option to compare against HIP-95, because this is only slightly worse on a sliding scale, in our opinion.

The impact of not proceeding with this HIP is a continuation of difficulties of onboarding for delinquent makers that have not decided if they are in the ecosystem for the long-haul nor have the ability in support their customers on a daily or even on a weekly bases. 

# Unresolved Questions
[unresolved]: #unresolved-questions

- Won't some select makers that already burned TONS of HNT for DC into their maker account be mad that the community is slashing the cost to onboard a hotspot by 10x, making it seem that they overpaid for DC?
  - Ultimately we would have to hear directly from those makers (Sensecap, Milesight, Browan/MerryIoT); as assuming they will be upset at this change doesn't change what we are trying to fix here.
  - Future economics HIPs can be devised to siphon off DC from maker accounts if those three deem it necessary as an 'overpayment' into the network. We personally don't think this is a problem, and is drop in the bucket for them as they are all out attending multiple conventions and conferences every month for business outside the Helium ecosystem. But regardless that can be a separate HIP. 

This unresolved question is for Alternative #0 (Amendment this this proposal) posted above:
  - What happens to maker account DC if we eliminate makers from paying any fees? 
    - Could be discussed and we have quite a few ideas, but out of the premise of this HIP to solve.  This would add additional complexities that would bog down the fix to the ever ongoing onboarding problem for all Hotspot Depoyers.  

# Deployment Impact
[deployment-impact]: #deployment-impact

If this HIP passes:
  - The onboarding fees should be set to $10 (1,000,000 DC) within 7 days. 
  - All makers will be informed on the change to $0 onboarding fee and the minimums required for a DC/SOL balance in their maker account, and instructed to make sure they keep tabs on their DC balance at regular interval as there is no official verbal warning from the Foundation or MCC member between 0-7 days of being in non-compliance. Recommended DC and SOL threshold levels will be told to each maker to account for 14 days of recent attempted/successful unique onboarding hotspots to get everyone significantly filled.

# Success Metrics
[success-metrics]: #success-metrics

If successful, new and current hotspot operators that are still deploying new hotspots will no longer have these pain points when out in the field setting up a hotspot for the first time.  And as such we will see a significant reduction of reports of the message "manufacturer undefined does not have enough SOL or Data credits messages" within Discord and other social media. Which has significantly has been hindered the IoT network from growing any further.


mrfizzy99:
Creating new.

PR #821, opened 2023-12-08
HIP 105: Modification of MOBILE subDAO Hex Limits

heatedlime:
This Helium Improvement Proposal (HIP) suggests modifying how Modeled Coverage Points (MCP) are calculated for Wi-Fi Access Points (APs) and CBRS Radios that provide overlapping coverage.


Ibra979:
Hi all .. can i have an indoor wifi hotspot and outdoor one in the same place ? .

PR #827, opened 2023-12-12
resolve stray includes

mawdegroot:
Resolving some of the stray include directories that the various HIPs have added to the repository.


abhay (approved):
Nice cleanup.

abhay:
One thing I'd like to do is to add a line to the updated HIPs here that state the "approved version" since we're making formatting updates. Just for posterity. 

Something like this https://github.com/helium/HIP/blob/main/0084-service-provider-hex-boosting.md?plain=1#L9

I think this can merge now though.

cc: @waveform06

PR #830, opened 2023-12-12
New HIP proposal - Finetune HIP17 parameters to tackle density

riobah:
Finetune HIP17 parameters to tackle density


waveform06:
There is a rendered view of this HIP with the attached images here
https://heliumgeek.com/0000-finetune-hip17-parameters-to-tackle-density.pdf

gradoj:
The proposed solution makes the scaling increase in many of the dense examples. This is the opposite of the hips stated objective to "focusing on addressing the increased hotspot density in urban areas." Would reducing hip15 redundancy variable not give desired results of less rewards overall to dense areas?

riobah:
As shown in the appendix images and on the live map, urban areas are generally scaled more significantly in this proposal. Regarding certain urban hexes like some in NYC receiving a higher scale, this is due to the relaxed scaling at the resolution-4 level, which previously caused sharp boundary edges. For example, hex `8a2a10728287fff` was primarily scaled at the res-4 level.

In HIP15, the redundancy variable, `N`, predominantly scales witness rewards (when `w > N`). Since most beacons are fully witnessed, reducing `N` is unlikely to decrease total reward units, as shown in the chart.
HIP15 reward distribution

gradoj:
Yes, removing r4 makes some dense areas worse. It is opposite of the stated goal of the hip so doesn't seem acceptable. 

I must be thinking about it wrong cause reducing redundancy from 4 to 1 would reduce the earnings across the board in dense areas - which aligns with the stated goal of the hip more than the proposed hip17 tweak

PR #846, opened 2023-12-21
Fix misprint in HIP-102 "disllowment" -> "disallowment"

slivingston:
This is trivial but worth fixing to support text search for the keyword "disallow".


abhay:
@jthiller clarification for you to check on.

jthiller (approved):
Thank you @slivingston

PR #857, opened 2023-12-31
HIP 107: Deterring Gaming Through Verification Fees

heatedlime:
This Helium Improvement Proposal (HIP) proposes requiring a fee, in HNT, be paid by deployers to the MOBILE subDAO Treasury to submit new Certified Professional Installer (CPI) information to Spectrum Access System (SAS) for outdoor CBRS Radios.


abhay:
I get the point of this in the short term (changing the denominator of the swap rate calculation such that the treasury floor bid is adjusted upwards) but I'm not sure if I agree with the long term dynamics and the short term volatility of the treasury curve (intraday va daily). I'd rather see a change in the utility score for the epoch through an HNT to DC burn or (as considered in the HIP) a reduction in MOBILE circulating supply. Or it'd be interesting to at least have an economic analysis of all three in the HIP.

heatedlime:
@abhay Updated to include Economic Impact Analysis section. By funding the treasury directly with HNT, it increases the MOBILE swap rate 0.00234992%, where as by re-adjusting the DAO utility score to count all HNT burn towards emissions, this only increases it by 0.00157036%.

I don't quite understand how this causes short term volatility of the treasury curve, so that might be a conversation to have offline. But based on my analysis, it makes the most sense to directly fund the MOBILE subDAO treasury with HNT fees.

PR #858, opened 2024-01-01
Create 0000-Preventing Rug Pull, Conflicts of Interest, and Centraliz…

raynef1re:
HIP101 written by a Nova employee proposes changes to the network that would start a cycle of transition away from CBRS by redistributing rewards away from the community members who invested in building out the CBRS network.  As stated in the HIP, the effect would be a long term reduction in rewards to CBRS.  What it doesn’t state is that the primary effect of that HIP would be a long term increase in profits for Nova Labs by pivoting to a technology that would earn Nova Labs more at the expense of the CBRS community.  This also disincentivizes the community from finding a fix for the CBRS issue they claimed was the reason for the HIP.  The conflict arises when Nova Labs employees can propose changes that undermine the investments of the community to the profit of their own company.

Proposals for network changes made by Nova Labs (previously known as Helium Inc) employees can give an impression of indirect centralization, potentially jeopardizing the network's decentralized status. This perceived proxy control over network decisions could attract the scrutiny of regulatory bodies such as the SEC, raising concerns about the network's compliance with decentralization standards. Therefore, to mitigate this risk and maintain the integrity of the network's decentralized nature, it is proposed that Nova Labs employees be excluded from the voting process on these matters.

By restricting Nova Labs employees from proposing changes to the network, this proposal supports the use case of maintaining a truly decentralized network. This is crucial for the network's integrity, as it prevents any single entity from exerting undue influence or control over the network's evolution.

Many community members have invested significant resources in building out the CBRS network. This proposal supports the use case of safeguarding these investments by preventing proposals that could undermine the value of these community-driven efforts, ensuring that their contributions remain valuable and relevant.

The proposal advocates for transparent and equitable governance within the Helium network. By making Nova Labs employees' wallets public and barring them from voting on HIPs, it supports the use case of fair and transparent decision-making processes, where conflicts of interest are minimized.

This proposal also supports the use case of encouraging and empowering the community to actively participate in problem-solving. By preventing Nova Labs employees from steering the network's direction, it opens up opportunities for community members to propose and implement solutions, fostering a more collaborative and innovative environment.

The proposal supports the crucial use case of maintaining regulatory compliance, particularly concerning decentralization standards that may be scrutinized by entities like the SEC. By minimizing direct involvement from Nova Labs, the proposal aims to uphold the decentralized nature of the network, thereby reducing the risk of regulatory challenges.

### DIFF
--- a/0000-Preventing Rug Pull, Conflicts of Interest, and Centralization
+++ b/0000-Preventing Rug Pull, Conflicts of Interest, and Centralization
@@ -1,0 +1,127 @@
+# HIP
+
+0000-Preventing Rug Pull, Conflicts of Interest, and Centralization
+
+- Author(s): @rf1re
+- Start Date: 1/1/2024
+- Category: meta
+- Original HIP PR: <!-- leave this empty; maintainer will fill in ID of this pull request -->
+- Tracking Issue: <!-- leave this empty; maintainer will create a discussion issue -->
+
+## Summary
+
+This proposal is to ban, prevent, or otherwise discard HIP proposals written by Nova Labs employees, as they have conflicting interest by putting the profits of a private company above the community members who build out the network.  This also proposes that employees of Nova Labs have their wallets made public, and barred from participating in voting on HIPs.
+
+## Motivation
+
+HIP101 written by a Nova employee proposes changes to the network that would start a cycle of transition away from CBRS by redistributing rewards away from the community members who invested in building out the CBRS network.  As stated in the HIP, the effect would be a long term reduction in rewards to CBRS.  What it doesn’t state is that the primary effect of that HIP would be a long term increase in profits for Nova Labs by pivoting to a technology that would earn Nova Labs more at the expense of the CBRS community.  This also disincentivizes the community from finding a fix for the CBRS issue they claimed was the reason for the HIP.  The conflict arises when Nova Labs employees can propose changes that undermine the investments of the community to the profit of their own company.
+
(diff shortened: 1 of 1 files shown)


heatedlime:
Looks like you have a duplicate PR, so posting this comment on both so it doesn't get lost:

This HIP doesn't really make sense. Just because a Nova labs employee puts a HIP forward, doesn't mean it will pass. The community votes on whether that happens.

Your HIP boils down to two things, which is actually not a problem -

1.) Nova labs employees should not write HIPs.
Pointless to have. I'll gladly put my name on any HIP a Nova employee recommends, or write one on behalf of their ideas. So this HIP would not solve that.

2.) Nova and it's employees should not be able to vote
Nova has stated that they as a company do not and has not voted with their wallet. Nova employees may have voted in the pass, but i doubt their vote was substantial enough to sway a vote. Employees of Nova are community members too, and want the network to succeed. Baring Nova employees votes won't change much IMO.

Further, you don't understand that no service provider wants to use CBRS, so why should we continue to reward CBRS at the same level as Wi-Fi? We have a service provider who is paying to offload data via wifi, but not CBRS (maybe $10 a day in CBRS offload vs $1700 a day in Wi-Fi offload). If you're so worried about the future of CBRS, why don't you go find a service provider who wants to join the network and use CBRS over Wi-Fi? Good luck, because the chances of winning the lottery are probably better.

abhay:
Which one of these two PRs do you want to be considered a draft @raynef1re?

waveform06:
This PR to be closed and the discussion continued in
https://github.com/helium/HIP/pull/859

PR #861, opened 2024-01-03
Create 0000-Suspend TMobile Partnership

raynef1re:


amirhaleem:
T-Mobile is not partnered with Helium, and the community does not get to vote on who we (Nova Labs) partner with as a private business for our own products.

abhay:
Although I would have done the same here, @amirhaleem please let the HIP Editors or the Foundation close inappropriate HIPs in the future.

PR #873, opened 2024-01-10
HIP 98: Updates

heatedlime:
Removed duplicative section and also added more details relating to averaging speed tests done in the same epoch to get final used.


waveform06 (approved):
Amendment confirmed

PR #890, opened 2024-01-28
Hip increase poc wifi

marcialc:
I believe HIP 101 had the right idea on increasing the WIFI radios rewards, it just needed to be more balanced in terms of rewards compared to the CBRS radios. I hope this new hip is what everyone wanted.


waveform06:
This is a copy of HIP 101, with the same original author and HIP content it appears.
And is thus invalid. A new HIP is requested. Please follow the HIP process correctly.
https://github.com/helium/HIP/blob/main/0007-managing-hip-process.md

hiptron:
Closing as stale

PR #906, opened 2024-02-10
HIP 107: Updates

heatedlime:
Addressed Waveforms two comments:

-Added additional stakeholder of MOBILE app developers
- Clarified what happens if CPI isn't reviewed within 2 business days.


zer0tweets:
I would recommend to remove the "priority fee" concept. Feels like it adds unnecessary complexity that will require CPIs to differentiate how they treat priority queue and regular queue. 

It also feels like this HIP blends a few things together - fees for CPI/hotspot and rules/process for black listing hotspots.

PR #907, opened 2024-02-10
Create 0110-data-incentives-with-usage-caps.md

zer0tweets:
Add data incentives with usage caps HIP PR and create discord channel, so we can start discussion with the community.

### DIFF
--- a/0110-data-incentives-with-usage-caps.md
+++ b/0110-data-incentives-with-usage-caps.md
@@ -1,0 +1,70 @@
+
+# Data Incentives with Usage Caps
+
+-   Author(s): @zer0tweets
+-   Start Date: 02/12/2024
+-   Category: Economic 
+-   Original HIP PR:
+-   Tracking Issue:
+-   Vote Requirements: 67% to pass
+    
+
+## Summary
+This HIP proposes a mechanism for service providers to offer select hotspot owners special incentive programs in the form of limited, capped payments to compensate said hotspot owners for voluntarily allowing data usage that would, otherwise, be unpaid or disallowed.
+
+## Motivation
+This mechanism would compensate hotspot for voluntarily enabling new use cases on the network and empower service providers to deploy bespoke, strategic incentives to enhance data usage on the network.
+
+## Stakeholders
+-   Hotspot owners
+-   Service providers
+- Mobile Hotspot Vendors
+    
+
+## Detailed Explanation
+There are a number of instances where service providers on the Helium Network may want to offer special data incentives to hotpots that the current mechanism of purchasing and accounting for data on the network doesn’t allow:
+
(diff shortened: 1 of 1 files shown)


heatedlime:
Instead of offering an ongoing data usage incentive for deployers to turn these features on, wouldn't it be cheaper, and easier to implement a referral method for deployers who have unpaid data and free wifi turned on? For example, if a subscriber signs up for HM through my splash page, why not just give me a $5 kickback or something? This also wouldn't require a HIP since it doesn't mess with tokenomics/emissions.

zer0tweets:
> Instead of offering an ongoing data usage incentive for deployers to turn these features on, wouldn't it be cheaper, and easier to implement a referral method for deployers who have unpaid data and free wifi turned on? For example, if a subscriber signs up for HM through my splash page, why not just give me a $5 kickback or something? This also wouldn't require a HIP since it doesn't mess with tokenomics/emissions.

Referral piece is a good idea, but IMO HM subscriber acquisition and marketing is slightly separate from the core goal of this HIP.

heatedlime:
> > Instead of offering an ongoing data usage incentive for deployers to turn these features on, wouldn't it be cheaper, and easier to implement a referral method for deployers who have unpaid data and free wifi turned on? For example, if a subscriber signs up for HM through my splash page, why not just give me a $5 kickback or something? This also wouldn't require a HIP since it doesn't mess with tokenomics/emissions.
> 
> Referral piece is a good idea, but IMO HM subscriber acquisition and marketing is slightly separate from the core goal of this HIP.

The goal is to encourage deployers to turn on free wifi and or unpaid data, no? If so, offering them a way to obtain additional revenue by turning It on, it should create enough incentive to have them on, without messing with emissions.

abhay:
I don't see why this is a HIP but maybe I'm not 100% understanding the goals. 

Is there a protocol/economic change being proposed? 

```
This HIP proposes a mechanism for service providers to offer select hotspot owners special incentive programs in the form of limited, capped payments to compensate said hotspot owners for voluntarily allowing data usage that would, otherwise, be unpaid or disallowed.
```

As currently written, this is a Service provider using their wallet to pay one or many Hotspot owner(s). This doesn't really involve a protocol change.

Unless the intention is to redirect one of the existing emissions categories (PoC, Data Transfer, etc.) towards this type of activity and have it directed by the SP.

abhay:
> Because for these "special data incentives" to work, we need to be able to set a reward data cap on a per hotspot basis. Us setting a reward data cap on per subscriber basis in HIP82, very much needed to be a HIP and was debated quite a bit... so I feel this is a HIP also.

Isn't that just a Service Provider and a Hotspot owner enter a contract (off chain) that says, roughly...

```
Hey, please check this box that make your Hotspot broadcast a new SSID. This SSID will allow people around to see and connect to your Hotspot and you will allow more data to flow on your Hotspot for non Helium Mobile subscribers. We'll account for it as a separate settlement service "feed" and it'll be tagged on chain as just a consumer of data. Payments will happen and Hotspot owners will be happy.
```
 The separate "incentive" you're talking about is just icing on the cake if folks decide to participate, right? Nothing coming from the protocol.

abhay:
What's the plan on this HIP @zer0tweets? Withdrawn since HM is already experimenting Free WiFi?

abhay:
Checking in again, @zer0tweets. Plan to close in a week of inactivity.

PR #921, opened 2024-02-28
HIP 115: Redefining the MOBILE Maker Approval Process

heatedlime:
This Helium Improvement Proposal (HIP) defines the approval process and requirements for Manufacturers of MOBILE Hotspots and Access Points (MOBILE Makers) to get a Maker Key.

### DIFF
--- a/0000-redefining-the-mobile-maker-approval-process.md
+++ b/0000-redefining-the-mobile-maker-approval-process.md
@@ -27,53 +27,90 @@ In the case of Bobcat and Nebra, both have produced a 5G Hotspot; however, neith
 
 ## Stakeholders:
 
-**MOBILE Makers** - Manufacturers wishing to create MOBILE hardware will be required to adhere to the requirements within this HIP.
-**Manufacturer Compliance Committee (MCC)** - The MCC will need to ensure the requirements of this HIP are met prior to awarding a MOBILE Maker a Maker Key. 
+- **MOBILE Makers** - Manufacturers wishing to create MOBILE hardware will be required to adhere to the requirements within this HIP.
+- **Manufacturer Compliance Committee (MCC)** - The MCC will need to ensure the requirements of this HIP are met prior to awarding a MOBILE Maker a Maker Key. 
 
 
 ## Detailed Explanation
-This HIP imposes the following requirements MOBILE Makers must satisfy prior to them being given a Maker Key: 
+This HIP imposes MOBILE Makers must create a new MOBILE Maker Escrow Wallet in which they will not have access to withdraw funds from. Additionally, MOBILE Makers must satisfy the four below requirements prior to being awarded a Maker Key. This HIP will classify MOBILE Makers into three (3) tiers.  Please note, approval of this HIP will serve as Nova Lab’s subDAO approval. As such, Nova Labs will have satisfied all four requirements to this HIP.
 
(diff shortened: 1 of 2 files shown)


waveform06:
Q1: What does Stripping of Maker Keys actually mean?
Q2: Are MOBILE Hotspots really CBRS Hotspots? As both CBRS and Wi-Fi are MOBILE Hotspots
Q3: Which Tier has Nova satisfied (Tier 2), What about FreedomFI have they satisfied (Tier1) or do they both get combined into a new Tier3? Currently they are separate wallets.
Q4: For Tier3 makers retiring what happens to the $4,000 worth of DC they must keep as a min balance to avoid slashing?

If there are 4 requirements for each tier better to number them rather than bullet them.

A Tier3 vendor needs 50M x $0.0036 and 2B DC = >$200K
A Tier2 vendor needs 50M x $0.0036 and 500M DC and $10K MOB = >$195K
A Tier3 vendor needs 50M x $0.0036 and 2.5B DC and $10K MOB = >$215K
Those no's right currently?

Currently Tier3 vendors (FreedomFi and Bobcat) onboard ($40) and assert ($10) using DC on both MOBILE and IOT networks from the same wallet. So currently supporting 100 onboards for both networks  (inc assert) would need 1B DC not 400M DC. 

For slashing,  better we work with UTC defined epochs not calendar days.
Eg if at the end of epoch x, minimum balances are not met then they have until say the end of x+2 to refill, otherwise at the end of x+2 they will be slashed.

heatedlime:
> Q1: What does Stripping of Maker Keys actually mean? Q2: Are MOBILE Hotspots really CBRS Hotspots? As both CBRS and Wi-Fi are MOBILE Hotspots Q3: Which Tier has Nova satisfied (Tier 2), What about FreedomFI have they satisfied (Tier1) or do they both get combined into a new Tier3? Currently they are separate wallets. Q4: For Tier3 makers retiring what happens to the $4,000 worth of DC they must keep as a min balance to avoid slashing?
> 
> If there are 4 requirements for each tier better to number them rather than bullet them.
> 
> A Tier3 vendor needs 50M x $0.0036 and 2B DC = >$200K A Tier2 vendor needs 50M x $0.0036 and 500M DC and $10K MOB = >$195K A Tier3 vendor needs 50M x $0.0036 and 2.5B DC and $10K MOB = >$215K Those no's right currently?
> 
> Currently Tier3 vendors (FreedomFi and Bobcat) onboard ($40) and assert ($10) using DC on both MOBILE and IOT networks from the same wallet. So currently supporting 100 onboards for both networks (inc assert) would need 1B DC not 400M DC.
> 
> For slashing, better we work with UTC defined epochs not calendar days. Eg if at the end of epoch x, minimum balances are not met then they have until say the end of x+2 to refill, otherwise at the end of x+2 they will be slashed.

Comments have been addressed in newest commit. 

Q1: Added Clarification
Q2: Defined that MOBILE Hotspots are MOBILE CBRS Hotspots
Q3: Specified that if a parent company has subsidiaries that are T2 and T3, they can create separate wallets.
Q4: Added reference that once the MOBILE Maker passes the retirement audit by the foundation that the smart contract for slashing is removed.

Comment 5: Added numbers instead of bullets
Comment 6: $10 is 1,000,000 DC. So $50 is 5,000,000. So 5,000,000 times 100 onboards is 500M DC. Let me know if my math is wrong. 

Comment 7: changed days to epochs

PR #927, opened 2024-03-05
HIP 114: Incentive Escrow Fund for Subscriber Referrals

zer0tweets:
Adding HIP to implement incentive escrow fund for subscriber referrals, per most recent discussion and approval by the mobile working group.

### DIFF
--- a/0114-incentive-escrow-fund-for-subscriber-referrals.md
+++ b/0114-incentive-escrow-fund-for-subscriber-referrals.md
@@ -1,13 +1,17 @@
-# Incentive Escrow Fund for Subscriber Referrals
+# HIP 114: Incentive Escrow Fund for Subscriber Referrals
 
-- Author(s): @zer0tweets
+- Author: [@zer0tweets](https://github.com/zer0tweets)
 - Start Date: 2024-03-12
 - Category: Economic, Technical
-- Original HIP PR:
-- Tracking Issue: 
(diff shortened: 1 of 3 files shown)


waveform06:
What happens when there are no un-used programmatic emissions from the service provider bucket or less than the 100% to match the numbers from multiple SPs?
How is the unused portion currently calculated?

abhay:
Some thoughts shared elsewhere but wanted to have it here too. I think this HIP needs a few more pieces until it's ready for a merge.

1. filename has a fun typo. 
2. file is missing a heading section
3. stakeholder section doesn't discuss impact to them
4. the example section is a good conversation topic but i'm not sure if it belongs in the HIP. we've found that it often sets expectations on what _will_ be build rather than what _could_ be built. maybe worth clarifying that.

zer0tweets:
> Some thoughts shared elsewhere but wanted to have it here too. I think this HIP needs a few more pieces until it's ready for a merge.
> 
> 1. filename has a fun typo.
> 2. file is missing a heading section
> 3. stakeholder section doesn't discuss impact to them
> 4. the example section is a good conversation topic but i'm not sure if it belongs in the HIP. we've found that it often sets expectations on what _will_ be build rather than what _could_ be built. maybe worth clarifying that.

Updated 1-3. I suggest we keep the example as it helps provide more concrete understanding of what this HIP could results in.

PR #937, opened 2024-03-06
Create 0xxx-hip-voting-once-per-month.md

TheHeroBrine422:
Quickly wrote this up. Not entirely sure it is good but wanted to get feedback before putting more effort into it.

### DIFF
--- a/0xxx-hip-voting-once-per-month.md
+++ b/0xxx-hip-voting-once-per-month.md
@@ -1,0 +1,44 @@
+# HIP Voting Once Per Month
+
+- Author(s): @TheHeroBrine422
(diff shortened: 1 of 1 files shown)


abhay:
I don't think this is a HIP exactly as it's more of a process that could be followed by the Helium Foundation to bring in HIPs and move them along in the governance process (getting a channel, preparing for a vote, voting, etc.) That said, I feel the pain here acutely.

abhay:
I'm going to close this HIP for now. I don't think it really makes sense for us to slow down to "once a month" for governance. I think there are other efforts going in flight to help with the fatigue that some may be having here.

PR #949, opened 2024-03-21
Create xxxx-modification-of-net-emissions.md

heatedlime:
This Helium Improvement Proposal (HIP) proposes changing net emissions whereby per epoch, 50% of HNT burned for Data Credits (DC) will be permanently removed from the max supply, while the remaining 50% will be reminted via net emissions for that epoch.


heatedlime:
Closing this for now. I think it's actually important that tokens are being re-minted and distributed mostly to the MOBILE subdao, since the treasury swap price is so low, this is effectively (for the time being) burning the HNT anyways.

PR #969, opened 2024-03-31
HIP 119: Closing Gaming Loopholes Within the MOBILE Network

heatedlime:
This Helium Improvement Proposal (HIP) proposes closing gaming loopholes to help prevent gaming within the MOBILE network.

### DIFF
--- a/0119-closing-gaming-loopholes-within-the-mobile-network.md
+++ b/0119-closing-gaming-loopholes-within-the-mobile-network.md
@@ -1,11 +1,13 @@
-# HIP XXX: Closing Gaming Loopholes Within the MOBILE Network
+# HIP 119: Closing Gaming Loopholes Within the MOBILE Network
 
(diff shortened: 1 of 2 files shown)


riobah:
Would it make sense to have the elevation limit in meters, as for true location to distance limits?

heatedlime:
> Would it make sense to have the elevation limit in meters, as for true location to distance limits?

So I think people in the US are used to feet, and can probably reasonably estimate the height of something in feet. That's probably not the same if you switch it to meters.

So then the question is should skyhook distance be switched to feet vs meters? Maybe, my guess is skyhook defaults the value to meters, so then a calculation would need to be done to turn it into feet, so it might take a little more work, but definitely not opposed to that. Let me ask this question to Boris and see what he thinks.

madninja:
Let’s please use meters. Converting to ft can be a UI  feature but everything works in meters under the hood

PR #975, opened 2024-04-05
HIP-000 - Slash Helium Mobile for incorrect RF pattern

AcknorAbyss:
Slash Helium Mobile for incorrect RF pattern on KP-3Q0MA-13


AcknorAbyss (commented):
/

abhay:
I think this is an important conversation to have and welcome it but slashing the service provider doesn't make sense, in my opinion. I know technically the owning entity is the same but Helium Mobile, as a Service Provider with 500m MOBILE staked, and Nova Labs/FreedomFi, as a Hotspot and gateway manufacturer 50m MOBILE staked, are two separate "entities" from a protocol perspective.

abhay:
Any thoughts here? It's been about 3 weeks since my last comment. Will likely close next week but appreciate the submission.

AcknorAbyss:
The only thing is that they are two separate "entities" from a protocol perspective in terms of staking and legal forms. Why both should be slashed has a heavier weight than just that.
There is no division of roles, responsibilities, and labor for mid-to-high salaried employees. From the C-suite to the coders, all have Nova and Helium Mobile hats. While the first-line customer support for Helium Mobile is outsourced. There is no mention anywhere publicly that certain employees are specific to either entity. For this mishap, if it came down to a single employee that works on both sides, are we going to ignore that they were doing work for both entities and the mishap happened on both entities due to this? It is mentioned in the document that planner.hellohelium.com has had this antenna for months. So who is responsible for the planner, Nova or Helium Mobile? If the answer is Nova which is doing work on Helium Mobiles' behalf with no checks and balances, then Helium Mobile should also bear the same responsibility for the error and misinformation on the planner.

Let the community decide.

abhay:
Although I understand the thinking here (not making a judgement on whether or not it's correct), I think it's important for the HIP process to be focused on **protocol level changes**. Some of these are social but most of them are technical. As it stands, the protocol does see Helium Mobile as a Service Provider and Nova Labs, the hotspot manufacturer who happens to make both IOT and MOBILE Hotspots, as different "entities".

If a correction is made to slash the correct entity I think this HIP would be fully formed for numbering and community discussion. I don't have any objections on the value to slash here. That's up to you, the author, to propose and the community to approve when it gets to that state in the governance process.

Until then, I'm good with the ongoing conversation as a pull request or in the community until there's clarity on the end goal here.

waveform06:
@AcknorAbyss 
There are discussions going on about this HIP in the #hip-discussion channel on the discord if you would like to join in

hiptron:
Closing due to lack of specificity and activity. Open to re-opening.

PR #984, opened 2024-04-17
Update 0115-redefining-the-mobile-maker-approval-process.md

heatedlime:
Changed requirements to be more aligned with foundation feedback.

### DIFF
--- a/0115-redefining-the-mobile-maker-approval-process.md
+++ b/0115-redefining-the-mobile-maker-approval-process.md
@@ -35,37 +35,48 @@ In the case of Bobcat and Nebra, both have produced a 5G Hotspot; however, neith
 
 
 ## Detailed Explanation
-This HIP imposes the Helium Foundation, on behalf of MOBILE Makers, to create a new MOBILE Maker Escrow Wallet in which the Maker will not have access to withdraw from. Additionally, MOBILE Makers must satisfy the five below requirements prior to being awarded a Maker Key. This HIP will classify MOBILE Makers into three (3) tiers.  Please note, approval of this HIP will serve as Nova Lab’s subDAO approval.
+This HIP imposes the Helium Foundation, on behalf of MOBILE Makers, to create a new MOBILE Maker Escrow Wallet in which the Maker will not have access to withdraw from. Additionally, MOBILE Makers must satisfy the below requirements prior to being awarded a Maker Key. This HIP will classify MOBILE Makers into three (3) tiers.  Please note, approval of this HIP will serve as Nova Lab’s subDAO approval.
 
 ### MOBILE CBRS Hotspots Only (Tier 3)
-1. Approval of hardware design by the Helium Foundation as outlined in HIP [HIP 19](https://github.com/helium/HIP/blob/main/0019-third-party-manufacturers.md)
-2. Approval from the MOBILE subDAO via veMOBILE vote with 67% approval
-3. The staking of 50M MOBILE tokens
-4. Provide a copy of the Makers secure boot key in an escrow wallet held by the foundation
-5. The Maker must have the following minimum balances within their Maker Wallet:
+1. Propose a HIP to become a MOBILE Maker
+2. Work with the Helium Foundation and Firmware developers to ensure their firmware is compatible
+3. Complete and publish penetration testing results of hardware
+4. Provide evidence to the Helium Foundation they have the source of liquidity to provide the required stake
+5. Their HIP goes to vote and approval from the MOBILE subDAO via veMOBILE is granted with at least 67% approval
+6. The Maker stakes of 50M MOBILE tokens
+7. The Maker provides a copy of the Makers secure boot key in an escrow wallet held by the foundation
+8. The Maker must have the following minimum balances within their Maker Wallet:
   - 1,000,000,000 DC
   - 2 SOL
   - 1 MOBILE
 
 ### Wi-Fi Access Points Only (Tier 2)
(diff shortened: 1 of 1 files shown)


0115-redefining-the-mobile-maker-approval-process.md
```diff
@@ -35,65 +35,77 @@ In the case of Bobcat and Nebra, both have produced a 5G Hotspot; however, neith
...
-1. Approval of hardware design by the Helium Foundation as outlined in HIP [HIP 19](https://github.com/helium/HIP/blob/main/0019-third-party-manufacturers.md)
-2. Approval from the MOBILE subDAO via veMOBILE vote with 67% approval
-3. The staking of 50M MOBILE tokens
-4. Provide a copy of the Makers secure boot key in an escrow wallet held by the foundation
-5. The Maker must have the following minimum balances within their Maker Wallet:
+1. Propose a HIP to become a MOBILE Maker
+2. Work with the Helium Foundation and Firmware developers to ensure their firmware is compatible
+3. Complete and publish penetration testing results of hardware
+4. Provide evidence to the Helium Foundation they have the source of liquidity to provide the required stake
+5. Their HIP goes to vote and approval from the MOBILE subDAO via veMOBILE is granted with at least 67% approval
```
abhay, line 45:
```suggestion
5. Their HIP goes to vote and approval from the MOBILE subDAO via veMOBILE is granted with the current thresholds required of MOBILE governance (today, 67% approval and 100,000,000 veMOBILE vote power)
```
modify to be something like this. feel free to wordsmith

0115-redefining-the-mobile-maker-approval-process.md
```diff
@@ -35,65 +35,77 @@ In the case of Bobcat and Nebra, both have produced a 5G Hotspot; however, neith
...
+6. The Maker stakes of 50M MOBILE tokens
```
abhay, line 46:
```suggestion
6. The Maker stakes a bond of MOBILE tokens (currently specified as 50,000,000 through HIP-53)
```
again more generic.

0115-redefining-the-mobile-maker-approval-process.md
```diff
@@ -35,65 +35,77 @@ In the case of Bobcat and Nebra, both have produced a 5G Hotspot; however, neith
...
+7. The Maker provides a copy of the Makers secure boot key in an escrow wallet held by the foundation
+8. The Maker must have the following minimum balances within their Maker Wallet:
   - 1,000,000,000 DC
   - 2 SOL
   - 1 MOBILE
```
abhay, line 51: do you want to add a USDC minimum here too?

0115-redefining-the-mobile-maker-approval-process.md
```diff
@@ -35,65 +35,77 @@ In the case of Bobcat and Nebra, both have produced a 5G Hotspot; however, neith
...
-2. Approval from the MOBILE subDAO via veMOBILE vote with 67% approval
-3. The staking of 50M MOBILE tokens
-4. Provide a copy of the Makers secure boot key in an escrow wallet held by the foundation
-5. The Maker must have the following minimum balances within their Maker Wallet:
+1. Propose a HIP to become a MOBILE Maker
+2. Work with the Helium Foundation and Firmware developers to ensure their firmware is compatible
+3. Complete and publish penetration testing results of hardware
+4. Provide evidence to the Helium Foundation they have the source of liquidity to provide the required stake
+5. Their HIP goes to vote and approval from the MOBILE subDAO via veMOBILE is granted with at least 67% approval
+6. The Maker stakes of 50M MOBILE tokens
```
abhay, line 59: see above.

0115-redefining-the-mobile-maker-approval-process.md
heatedlime: Since onboard fee's for CBRS hotspots are solely paid in DC, I don't think we need to have any USDC requirements.

abhay: One thing that this would force us to do is to properly build/audit/use the USDC -> HNT -> DC automatic process we've been building already. 

I think, in general, it makes sense that people that need to burn something on chain want to hold things in USDC rather than in whatever assets are required. (HNT/DC, MOBILE, IOT, etc.) Making it optional (either USDC in escrow or DC) is the right direction to support these "payers" on the network.

heatedlime: There's also the low risk that USDC depegs again from $1.00 in a black swan event, similar to FTX collapse. I think it should be encouraged to hold DC for the DC portion rather than USDC.

waveform06: I believe there are many reasons why IOT vendors dont keep their DC balance topped up to a decent amount, but these two are important.
(1) DC is one way, burn HNT to DC and you can never get it back unless you can delegate it to an LNS.
(2) Wallet administration. Over the years vendors have had changes in personnel. Anyone touching the wallet could have written down those keys (inc ex personnel) so they dont want much transferable tokens in wallets (ready to burn to DC) that ex employees could have access to the funds. (New wallets is a possibility here - when DC runs out).
Having a USDC>HNT>DC process they dont touch helps with their token security. And then adding USD>USDC Eskrow means accessibility of transferable tokens with less people that touch the device that has the keys.

waveform06: I'll leave it up to Abhay or Scott - but maybe its a request for $10,000 worth of (tokens in any combination of USDC, HNT and DC)?

abhay: It does not complicate it. Nope.

PR #985, opened 2024-04-18
HIP 111: Updates

capjbadger007:
Updated text and unrewarded data added to be counted for the benchmark.

### DIFF
--- a/0111-mobile-data-utility-benchmark.md
+++ b/0111-mobile-data-utility-benchmark.md
@@ -11,11 +11,11 @@
  
 ## Summary
 
-This Helium Improvement Proposal (HIP) introduces data transfer requirements for MOBILE CBRS Radios/Wi-Fi Access Points, hereinafter referred to as Radios/APs, to be eligible for Boosted Coverage Points (CPs) in a given epoch. Units meeting the specified data transfer benchmark (set initially to 1 MB within the previous 7 epochs) will receive 10x CPs for the current epoch. The boost in Coverage Points will result in increased MOBILE rewards for those Radios/APs. 
+This Helium Improvement Proposal (HIP) introduces data transfer requirements for MOBILE CBRS Radios/Wi-Fi Access Points, hereinafter referred to as Radios/APs, to be eligible for Boosted Coverage Points (CPs) in a given epoch. Units meeting the specified data transfer benchmark (set initially to 1 MB within the previous 7 epochs) will receive a multiplier to their CPs for the current epoch. The boost in Coverage Points will result in increased MOBILE rewards for those Radios/APs. 
 
 ## Motivation
  
-To ensure a fair and active participation of MOBILE Radios/APs in the Helium Network, it is essential to establish benchmarks that encourage consistent data transfer through these units. This proposal addresses the need for a measurable metric to gauge the activity of MOBILE units, enhancing the overall health and performance of the network.
+To ensure a fair and active participation of MOBILE Radios/APs in the Helium Network, it is essential to establish benchmarks that encourage consistent data transfer through these units as a measure of the utility of the installation. This proposal addresses the need for a measurable metric to gauge the activity of MOBILE units, enhancing the overall health and performance of the network.
 
 ## Stakeholders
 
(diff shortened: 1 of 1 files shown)


heatedlime:
The following are drawbacks of this HIP, but aren't noted in the drawback section:

1.) Creating a weekly data requirement creates an environment that encourages deployers to deploy these in areas close to home, regardless of if the location is a good location or not and discourages people from even seeking out locations that are far away from them. 

2.) This HIP encourages inorganic data being passed through radios, which comes at the expense of Service Providers who are actually paying for the data. When more inorganic data is passed, the amount of PoC rewards for everyone decreases, which is a drawback to every deployment.

3.) Unless a map of Subscriber data is released, there's no tools for deployers to know where to deploy these that may pass data. Which will then result in the deployers from having to go map their deployments once a week.

capjbadger007:
> 1.) Creating a weekly data requirement creates an environment that encourages deployers to deploy these in areas close to home, regardless of if the location is a good location or not and discourages people from even seeking out locations that are far away from them.

Not at all. It encourages installs in areas where there is actual usage. 

> 
> 2.) This HIP encourages inorganic data being passed through radios, which comes at the expense of Service Providers who are actually paying for the data. When more inorganic data is passed, the amount of PoC rewards for everyone decreases, which is a drawback to every deployment.

This is also not true.

> 
> 3.) Unless a map of Subscriber data is released, there's no tools for deployers to know where to deploy these that may pass data. Which will then result in the deployers from having to go map their deployments once a week.

Deployers will not have to pass data through their own radios unless they have installed in areas where there are no people to use them.

> 
> 4.) New deployments that are passing tons of data will be at a disadvantage in terms of PoC rewards for the first 60 days.
You have misread the HIP. The Ramp up is a gradual system for the multiplier for everyone. It's not a ramp up for each install.

heatedlime:
1.) How are deployers suppose to know areas that will provide actual usage if there's no subscriber map available? 

2.) This is true, and you didn't even try to refute it. There's literally data/tokenomics/emissions schedule that backs it up. 

3.) See comment 1

4.) the HIP needs to be written better to state this.

heatedlime:
Screen Shot 2024-04-18 at 11 17 42 AM
More than half of Wi-Fi deployments that touch a High Footfall hex have 0 data passed within the past 7 days. How are deployers supposed to know where to deploy these if there's no subscriber map published?

capjbadger007:
Why are they not passing data? Should focus on that rather than using them as a poor excuse to be against the network being used.

heatedlime:
> Why are they not passing data? Should focus on that rather than using them as a poor excuse to be against the network being used.

You really are helpless. I encourage the foundation to not merge this PR until the drawbacks are put into the HIP.

capjbadger007:
You have stated that your hotspots don't pass data. What use are they to the network? That is your objection and always has. We want a healthy network.

heatedlime:
> You have stated that your hotspots don't pass data. What use are they to the network? That is your objection and always has. We want a healthy network.

Where should i move it to?

capjbadger007:
> 
> FWIW, I don't think that "this will be implemented in 30 days" is a plan of implementation. But we can continue to iterate on the HIP until then.

It's the same thing that was in 113. 
It was meant to be a minimum 30 day gap after voting to allow coms to make sure people are aware of it. I'll reword it.

heatedlime:
> > You really are helpless. I encourage the foundation to not merge this PR until the drawbacks are put into the HIP.
> 
> I disagree with this way of dealing with a proposal you don't agree with. To be clear, the Foundation's role is to make sure community proposals come to a vote if they reach a level of discussion that should likely be opened up to the entire network to decide. If the author feels like this is ready for that wider conversation and there's clarity on implementation, I'm okay with this coming to a vote.
> 
> FWIW, I don't think that "this will be implemented in 30 days" is a plan of implementation. But we can continue to iterate on the HIP until then.

So if the author refuses to add valid drawbacks of the HIP just so voters aren’t informed, the foundation is ok with that?

abhay:
> So if the author refuses to add valid drawbacks of the HIP just so voters aren’t informed, the foundation is ok with that?

It is _okay_ for HIP authors to decide what's in their HIP and it's also okay for the community to vote on the merits of the HIP.

PR #986, opened 2024-04-18
HIP 120: MOBILE Data Connectivity and Usability

mrfizzy99:
This Helium Improvement Proposal (HIP) introduces Data Connectivity and Usability requirements for MOBILE CBRS Radios and Wi-Fi Access Points in which units must be rewarded 1 data credit over a 14 epochs (days) in order to receive 1.0x multiplier for mobile coverage points. Units that do not pass this metric will receive 0.5x multiplier for mobile coverage points.


abhay:
Is this intended to be an alternative to HIP 111? Are you removing yourself as an author of HIP 111 as well if this is an alternative? Happy to consider merging this after some review. Going to have a look in a bit after MOBILE call.

mrfizzy99:
Yes;  Sure, the crux is that I can't remove something I can not edit.

To much hate on hip-111 over something silly, so remove that variable all together and see where the hate redirects to.

abhay:
Understood. I'll remove you as an author on the other HIP as well to make that clear that this is a separate HIP and not a modification.

abhay:
https://gist.github.com/abhay/1f6b4b8ce0efd7934ad0051b28b7e1e1/revisions

Some formatting and typo fixes here that i'd recommend as well. Looks like you're submitting some NBSPs which github and other environments like to call out. Can you clean those up in the future too?

PR #991, opened 2024-04-25
HIP 121: Service Provider Hex Boosting Improvements

jaym2518:
A HIP to address issues that have been identified in HIP-84: Service Provider Hex Boosting. The proposed changes include the introduction of a cap to the multiplier allowed to be utilized by a Service Provider.

### DIFF
--- a/Create HIP XXXX: Service Provider Hex Boosting Improvements
+++ b/Create HIP XXXX: Service Provider Hex Boosting Improvements
@@ -1,6 +1,6 @@
 # HIP xxxx: Service Provider Hex Boosting Improvements
 
-- Author(s): [Jeremy Mahrle](https://github.com/heatedlime)
+- Author(s): [Jeremy Mahrle](https://github.com/jaym2518)
 - Start Date: 2024-04-25
 - Category: Economic, Technical
 - Original HIP PR: 
--- a/Create HIP XXXX: Service Provider Hex Boosting Improvements
+++ b/Create HIP XXXX: Service Provider Hex Boosting Improvements
@@ -35,9 +35,7 @@ This action will also mitigate the effects of a potential future gaming vector w
 
 To address these vectors, this HIP proposes creating a chain variable to allow for an upper cap to be placed on the multiplier that a SP can boost a hex. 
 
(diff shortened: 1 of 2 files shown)


Create HIP XXXX: Service Provider Hex Boosting Improvements
```diff
@@ -0,0 +1,57 @@
...
+This HIP makes the following change:
+
+**Cap on Boost Multiplier: ** We have learned that boosting hexes at too large of a percentage can encourage gaming behavior in order to collect increased PoC rewards, such as spoofing GPS coordinates, speed test results, and/or the Skyhook location of a deployment. By capping the amount that a SP can boost, we can reduce instances of this gaming vector.
+
+This action will also mitigate the effects of a potential future gaming vector where a SP could boost hexes where their own deployments are located to such an extent that they would be able to collect nearly all the PoC rewards issued in an epoch.
+
+To address these vectors, this HIP proposes creating a chain variable to allow for an upper cap to be placed on the multiplier that a SP can boost a hex. 
+
+Initially, this variable will be set to 20x. In the future if there is a desire to change this variable to a different cap, the change may be proposed in two ways: 
+1)	This HIP authorizes the Mobile Working Group, upon a majority vote of its members, to put forward a governance proposal to veMOBILE voters for their approval, meaning a new, formal HIP is not required.
```
abhay, line 39: Although I agree that this is potentially a faster process, I think it's important to always have a HIP for every protocol change. I'd be fine with a request for this numbered proposal to be fast tracked in all cases to a vote. 

Maybe a template could be written for this.

jaym2518: Would this work?

"Initially, the variable will be set to 20x. In the future if there is a desire to change the variable to a different cap, a new HIP will be required. If HIP authors follow the standard HIP template and limit their proposal to only adjusting this specific chain var with no other proposed changes, they will be able to request that the Foundation fast-track the proposal to a vote."

abhay: Feels reasonable.

jaym2518: Submitted this as a commit (think I am using the phrasing correctly)

PR #1000, opened 2024-04-29
HIP 111: Remove author

abhay:
Removing author, as recommended in https://github.com/helium/HIP/pull/986#issuecomment-2078346006

@mrfizzy99 FYI. Let me know if this is what you were hoping to do with the opening of HIP 120.


mrfizzy99:
Confirmed. 

I liked and contributed to 111, but got to complex as we ingested ideas with peoples input. Then when we were unwilling to edit the simple data requirement, everyone got upset.
So offering an alternative koolaid, or they can go thirsty voting no on both and keep the current statuesque.

abhay:
I think it'll be interesting if both end up going to a vote. We need a way to reconcile if both pass.

PR #1006, opened 2024-05-07
HIP 123: Redefining IOT Data-Only Hotspot Onboarding and Assertion Fees

Ducktatorrr:
This first draft of the HIP introduces changes to the onboarding and location assertion fee structure for data-only IoT hotspots on the Helium network. The proposal aims to lower the onboarding fees to 500,000 DC ($5) and allow one free location assertion every 30 epochs to encourage wider adoption and more frequent updates of hotspot locations... or at least, that's what's in this first draft. 

This HIP was initiated by members of the IOT Working Group.


waveform06:
HIP Editor notes:
## Stakeholders: Need to add 
Developers of Helium Wallets or Maker Apps that allow location assertion
Developers of Helium CLI and API tools that allow location assertion

## Detailed Explanation.
Confirm that the first assert is free.

## Drawbacks
A theoretical possible one is that a Maker doesn't update their maker app and either DO onboards/asserts don't work, or they go thru at the wrong cost. But as the only DO vendor I know is Seeed and they will  almost certainly update the maker app, its probably not a real problem.

## Deployment Impact
Who is doing the blockchain work? Need to request Foundation/Nova to do this unless you have already got a resource.
And there is an impact on Developers of Helium Wallets, Maker Apps, CLI and API tools that allow location assertion.

"This proposal is fully reversible and compatible with existing network practices." doesn't really say anything.

## Sucess metrics
Is that an increase in Data-Only assertions or any location assertions?
And unless there is a metric for reflecting "greater network utilization and coverage." best leave it out as a metric.

Last sentence is probably better as last sentence in motivation section

PR #1042, opened 2024-06-13
remove planner reference

madninja:
Planner will indeed need to reflect that boost rewards could be lower than the multiplier indicates, if it doesn’t already. This is not a HIP detail though, so removing it here for clarity


mrfizzy99:
Approved from HIP author.

PR #1088, opened 2024-09-19
Add IOT data pricing HIP

rawrmaan:
Adds proposal to increase data pricing on the IOT network from `1 DC` to `50 DC` per packet.

PR #1089, opened 2024-09-19
better iot pricing hip

BFGNeil:
a better look at iot pricing with use case explaining why 10x is the right figure.


abhay:
@BFGNeil how close are we to landing this one or the other one with @rawrmaan #1088

PR #1091, opened 2024-09-23
HIP 134: Reward MOBILE Carrier Beta Hotspots

jhella:
This was drafted based on discussion during the 9/19 MWG Meeting.


WorkingTitle-Helium:
If there's a way to know how many connections a hotspot gets AND if that can be a variable used to modify rewards a simple modifier based on connections is cleaner I think. 

"A hotspot serving unique connections will receive an additional 0.01 modifier per connection up to 100 connections." That way connections are the incentive, no need to track paid or unpaid data. Goal is to incentivize new hotspots being deployed in high traffic areas. You can drop the bump to 1.0 for hotspots in 103 modified areas as unique connections will drive what they should  be rewarded. Cap it at 100, either daily or rolling weekly, whichever is simpler.

jhella:
> If there's a way to know how many connections a hotspot gets AND if that can be a variable used to modify rewards a simple modifier based on connections is cleaner I think.
> 
> "A hotspot serving unique connections will receive an additional 0.01 modifier per connection up to 100 connections." That way connections are the incentive, no need to track paid or unpaid data. Goal is to incentivize new hotspots being deployed in high traffic areas. You can drop the bump to 1.0 for hotspots in 103 modified areas as unique connections will drive what they should be rewarded. Cap it at 100, either daily or rolling weekly, whichever is simpler.

Nova knows how many connections a hotspot is getting, but need to discuss with them how easy is it would be to use it as a variable like you suggested.  Was hoping to get some feedback on this in the next MWG.

I like the idea of 0.01 modifier up to 100 connections if it's easy to implement.  I do think data transferred is important.  I could set-up my hotspot on top of a busy highway and get a lot of connections, but I probably won't transfer a lot of data.  We want high footfall areas with long dwell times.  Plus rewarded data means someone paid for it and results in HNT burn which is good for the ecosystem flywheel.

WorkingTitle-Helium:
Easy to implement is something I have no information on. Up to the devs to let you know. I’d offer run the modified subscriber multiplier the following epoch. I.e. Monday’s connections modify Tuesday rewards just to smooth compute cycles. 

Requiring data thresholds gives an incentive to artificially use carrier plans to get rewards. Consider the what that will look like.

WorkingTitle-Helium:
Back to the importance of number of connections visibility, if it's possible for that number to be shared and even used in rewards calculations. 

Just had a conversation with a host asking if I could tell them how many unique phones connected, seems they have an interest in near real time hard data about foot traffic.

madninja:
> Back to the importance of number of connections visibility, if it's possible for that number to be shared and even used in rewards calculations.
> 
> Just had a conversation with a host asking if I could tell them how many unique phones connected, seems they have an interest in near real time hard data about foot traffic.

Remember that carriers don't pay for unique connections, but pay for data over connections, so rewarding per connection may not be the right way to incentivize deployments. I suggest that unique connections are useful from a planning perspective (and in fact planner has a layer showing just that)  and could count as an eligibility criteria for rewards instead.

zer0tweets:
@WorkingTitle-Helium I am concerned that introducing some linear rewards accelerator as a function of connections / data served could be a recipe for gaming. My suggestion is to keep it binary. I.e. we agree on a threshold, like 25 unique connections in the last day and, if you met the threshold - you are excused from CDR and maybe some other requirements. If you didn't - then you are not.

WorkingTitle-Helium:
@zer0tweets My thoughts were around carrier offload is something Nova could disable if a hotspot was somehow selected and gaming, just update the hotspot in question to a non-carrier version and all the 131 requirements are in place. Also my inclination is hotspots in high traffic areas that get 100+ connections a day should be rewarded more than non-carrier hotspots and hotspots with even 25 connections. I see that as encouraging the community to deploy in commercial locations. I am not a fan of any data transfer benchmark as most data is now paid and that's enough of a reward. Looking forward to discussing this in a HIP channel.

PR #1095, opened 2024-10-05
HIP 135: Transitioning to Templated MOBILE Coverage

jhbr821:


zer0tweets:
An important point that this HIP doesn't make very explicit (and I think a lot of people misunderstand) is that base coverage points != actual number of reward points a radio will end up earning. The proposal is that points will be calculated using the same method as we had before with propagation API active, except now we'll be using an "assumed template" for each radio vs. clutter adjusted data from the propagation API. I.e. whether hexes overlap with other radio hexes, what HIP 103 multiplier a hex has and what boost multiplier a hex has still applies. TL;DR, radios don't actually have "point assigned", but rather a template with a number of high-medium-low signal strength hexes they cover is assigned. And then the actual number of points is calculated based on what this template covers. We should probably publish a graphic with pictures of templates and how they compare for different radios to make it more clear.

abhay:
made some direct recommendations as commits to reduce the back and forth @jhbr821 @madninja @zer0tweets

PR #1103, opened 2024-10-11
Fix Outdoor Wifi sample picture

jhbr821:


abhay:
Is this ready to land @jhbr821 ?

jhbr821:
Yeah I had just copied the wrong picture for WiFi outdoor so this was fixing that. Otherwise it is done from my perspective.

abhay:
Sounds good. We'll land it.

PR #1140, opened 2024-12-13
Create 0000_removal_of_service_provider_emissions.md

heatedlime:
This HIP would remove the Service Provider MOBILE reward pool/emissions that are currently given to Service Providers (SP) as a rebate for using the network.


heatedlime:
Icing for now

PR #1150, opened 2025-01-17
HIP 142 HeliumXchange

HeliumXXX:


waveform06:
This looks like a presentation for investment rather than a Helium Improvement Proposal.
You don't need a HIP to create this. HIPs are for direct changes to the network.
The Community does not vote Nova/Foundation to award grants.

This is described as a standalone dapp that interfaces with Helium NFTs on Solana.
Just build it, like other apps that interface with NFTs on Solana.
Or apply for a grant direct
https://www.helium.foundation/grants

PR #1158, opened 2025-02-12
clarify that HIP-143 applies to MVNO data, not MNO offload

ferebee:
The pricing mechanism of HIP-143 makes sense for MVNOs, but different considerations apply to Wi-Fi offload by MNOs. Although MNOs are already testing offload onto the Helium network in the US, it’s not publicly known how their data will be priced in production.

If this HIP doesn’t intend to define pricing for MNOs, that should be made explicit.

### DIFF
--- a/0143-multi-geography-mobile-economics-update.md
+++ b/0143-multi-geography-mobile-economics-update.md
@@ -11,7 +11,7 @@
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


ferebee:
Well, I guess this may all be quite irrelevant actually...

PR #1195, opened 2026-01-04
HIP 0149: FreedomFi Gateway Bootable USB Recovery and Custom OS Support

wicorn29:


waveform06:
Nova has withdrawn support for FreedomFi.
So no HIP will be voted on or implemented by Helium / Nova.
This could be implemented by independent developers and requires no HIP.

PR #1205, opened 2026-06-08
Update HIP references for Mobile and IoT PoC removal

waveform06:
Removed obsolete HIPs from Mobile and IoT PoC removal sections and updated references accordingly. Removed HIPs never went to vote or were rejected by a vote.


waveform06:
Should hex boosting HIPs also be revoked? 84,109,122,125

madninja:
Verified the cleanup against the README and the tracking issues. The 16 removed HIPs break down as:

- **Closed** (12): 21, 42, 44, 50, 57, 61, 62, 66, 94, 111, 127, 132
- **Rejected** (3): 72, 101, 107
- **Approved then Repealed** (1): 118

The 15 Closed/Rejected cases match the PR description's rationale cleanly. HIP 118 is the edge case worth noting for the record: it was Approved (its tracking issue carries the `approved` label) but later Repealed (also carries the `repealed` label) without ever being implemented — HIP 131's text says "HIP-118 is not yet implemented" and was structured as a bridge until 118 shipped, which it never did. So removing it from HIP 149's "retired by Mobile PoC removal" list is correct (HIP 149 isn't retiring anything still in force there); the reason just isn't "never went to vote or rejected", it's "repealed before implementation".

Retained list spot-checked: 131, 133, 135 are Deployed and 119 is Approved — active PoC-side mechanisms that HIP 149 would actually retire. Good cut.

Merging.

PR #1212, opened 2026-06-12
HIP 149: rework Advisory Council (compensation, voting, scope, timing)

madninja:
Reworks the Advisory Council (Decision 4) and reconciles it onto the latest `main` (#1213 terminology, #1215 roles doc).

## Compensation
- Paid only to the 5 community-nominated seats; the 2 Nova-appointed seats serve unpaid.
- 8,000 HNT per member per month, paid from the supplement vault (a ~1% charge on the supplement, not the Operations Funds).
- $10,000/member/month value cap; set by the authorizing vote (not deferred to first seating); adjustable forward by community vote (reduce at simple majority, increase at 66.67%). No discretionary fund.

## Voting
- Authorizing vote raised to a 66.67% supermajority (precedent HIP 85 / HIP 124); a proposal should not authorize its own issuance at a self-set simple-majority bar.
- Council-raised amend/terminate stays simple majority, downward only; any increase to issuance (rate, windows, compensation) takes the 66.67% bar.

## Scope
- Reframed as oversight, not control: no on-chain veto, no power to direct or block any individual spend. New spending requirements take effect only by community vote, not Council fiat.

## Timing
- M0 = Council seated; the program upgrade ships then. Mint starts M0 + 2 weeks, giving the seated Council a review window before issuance begins. Window boundaries anchor to mint start.

## Also
- Nova voting position and the 5-day Council vote-notice rule.
- USD reference reconciled to a ~$0.30 illustrative basis; HNT amounts are the binding commitment.
- Advisory Council roles doc updated in lockstep (compensation source, term, amendable parameters) and terminology aligned to target-minimum / payer rate.

### DIFF
--- a/0149-helium-utility-and-emissions-realignment.md
+++ b/0149-helium-utility-and-emissions-realignment.md
@@ -14,10 +14,10 @@ reddit-post-id: 1twsn06
 
 ## Summary
 
-This HIP bundles four governance decisions, executed in a single program upgrade at passage:
+This HIP bundles four governance decisions, executed in a single program upgrade that ships once the Council is seated:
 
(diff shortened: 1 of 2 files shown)


jaym2518:
5 days seems excessive. 3 days would be more than sufficient.

PR #1220, opened 2026-06-24
Strengthen non-compete clause to match CPA; Markdown formatting fixes

ferebee:


madninja (approved):
Matches the CPA's total-prohibition non-compete and fixes the tilde rendering issue. Thanks for the catch.

PR #1240, opened 2026-08-22
Foundation winding down IoT Network

gradoj:
@Scottsigel

### DIFF
--- a/IoT Operations Fund Transparency and Stewardship.md
+++ b/IoT Operations Fund Transparency and Stewardship.md
@@ -1,0 +1,228 @@
+# HIP: IoT Operations Fund Transparency and Stewardship
+
+- **Author(s):** @gradoj
+- **Start Date:** 2026-08-22
+- **Category:** Governance, Economic
+- **Original HIP PR:** TBD
+- **Tracking Issue:** TBD
+- **Vote Requirements:** veHNT Holders
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


leogaggl:
I wish you good luck! 

Neither the Helium Foundation nor @scottsigel, in particular, has any track record of transparency. And it appears that they have already jumped ship. The HF website seems to have been offline for weeks.

Most likely already abandoned before any announcements - same as per past record on employees.

gradoj:
If it was a legit nonprofit this transparency would already be available. 

What I don't think Scott realized before becoming a Nova yes man is that he is personal responsible for what happens at the non profit foundation. He can't turn off the website and give the funds to Nova those are legally the iot networks.

He's either very brave or very stupid.

leogaggl:
Unfortunately, this was always by design. The Foundation was created after Helium Inc and was the 'fig-leaf' designed to give the impression of some independence without ever doing anything real. Hot air was all they produced, really.

They hand-picked the people that couldn't think for themselves. And the track record speaks for itself. A litany of dodgy deals done behind closed doors, both by HF and Nova (and whatever they called themselves before) - waved through by a horde of happy clappers.

Not that the IOT demise should surprise anyone. This direction was clear when they pushed through the HIP51-53 and HIP70, and it was always going to end badly for LoRaWAN, as the economics were plucked out of the air in the early days and driven by drone speculators, not the builders, and they were never serious about reining in gambling.

